PR #2, opened 2017-05-25
Wait for browser switch result before clearing the request

quinnjn:
Firefox for Android with Tab Queue enabled backgrounds URL requests.

This pattern breaks Browser Switch:

- Browser Switch starts, the user decides to use Firefox Tab Queue to open the URL.
- Tab Queue backgrounds the new URL request, finishes, and Android foregrounds the caller app.
- BrowserSwitchFragment is resumed, and assumes the Browser Switch is complete.
- BrowserSwitchFragment get the incorrect result from BrowserSwitchActivity and clears the request.

The solution is for BrowserSwitchActivity to enter a new state, waiting for a Browser Switch response.
BraintreeSwitchFragment waits for BrowserSwitchActivity to prepare the proper response before
accepting the response, and clearing the request.

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
+++ b/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
@@ -34,6 +34,14 @@ public class BrowserSwitchActivity extends Activity {
         return sReturnUri;
     }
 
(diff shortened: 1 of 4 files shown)


quinnjn:
Unfortunately, trying to support Firefox Tab Queues would end up breaking the cancel responses. We'll have to close this PR for now until we can come up with some better idea.

PR #9, opened 2018-06-19
Move to support fragments

quinnjn:
Android P will deprecate `android.app.Fragment`. This PR adds support for the support fragment `android.support.v4.app.Fragment`

This PR includes
* Updating to SDK 27
* BrowserSwitchSupportFragment which has the same functionality as BrowserSwitchFragment, just inside of a support fragment
* Deprecating BrowserSwitchFragment
* BrowserSwitch package-private class which holds shared logic between the two fragments

### DIFF
--- a/browser-switch/src/test/java/com/braintreepayments/browserswitch/test/TestFragmentActivity.java
+++ b/browser-switch/src/test/java/com/braintreepayments/browserswitch/test/TestFragmentActivity.java
@@ -2,5 +2,5 @@ package com.braintreepayments.browserswitch.test;
 
 import android.app.Activity;
 
-public class TestActivity extends Activity {
+public class TestFragmentActivity extends FragmentActivity {
 }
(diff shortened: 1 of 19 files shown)


browser-switch/src/test/java/com/braintreepayments/browserswitch/test/TestFragmentActivity.java
```diff
@@ -0,0 +1,6 @@
+package com.braintreepayments.browserswitch.test;
+
+import android.support.v4.app.FragmentActivity;
+
+public class TestFragmentActivity extends FragmentActivity {
```
demerino: Maybe helpful to add a comment here that this new class is used for tests and is intentionally left empty?

quinnjn: Hmm we could, but it would only be visible for folks currently developing. 

It's in the `src/test/` directory and those files don't get published.

quinnjn:
Closing in favor of #10

PR #13, opened 2018-11-14
Doc singleInstance activities and browser-switch

quinnjn:
Issue reported here: https://github.com/braintree/braintree_android/issues/236

Activities that have a `singleInstance` launch mode (i.e. Activity launching on its own activity stack) will start Activities on a different activity stack which prevents Browser Switch from returning a result.


quinnjn:
@demerino unfortunately not. We don't have an programatic way to check that the Activity that BrowserSwitch is attaching to is not singleInstance.

PR #16, opened 2019-02-15
Add bundle support

callingmedic911:
We can use custom start activity animation using this.
```
    public void browserSwitch(int requestCode, Intent intent, Bundle startAnimationBundle) {...}
```

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
+++ b/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
@@ -9,6 +9,7 @@ import android.os.Bundle;
 import java.util.List;
 
 import androidx.annotation.Nullable;
+import androidx.core.content.ContextCompat;
 import androidx.fragment.app.Fragment;
 
 /**
@@ -116,6 +117,18 @@ public abstract class BrowserSwitchFragment extends Fragment {
      * @param intent an {@link Intent} containing a url to open.
      */
     public void browserSwitch(int requestCode, Intent intent) {
+        browserSwitch(requestCode, intent, null);
+    }
+
+    /**
+     * Open a browser or <a href="https://developer.chrome.com/multidevice/android/customtabs">Chrome Custom Tab</a>
+     * with the given intent and custom start activity animation.
+     *
+     * @param requestCode the request code used to differentiate requests from one another.
+     * @param intent an {@link Intent} containing a url to open.
+     * @param startAnimationBundle a {@link Bundle} for defining start activity animation.
+     */
+    public void browserSwitch(int requestCode, Intent intent, Bundle startAnimationBundle) {
         if (requestCode == Integer.MIN_VALUE) {
             BrowserSwitchResult result = BrowserSwitchResult.ERROR
                     .setErrorMessage("Request code cannot be Integer.MIN_VALUE");
@@ -140,7 +153,7 @@ public abstract class BrowserSwitchFragment extends Fragment {
         }
 
         mRequestCode = requestCode;
-        mContext.startActivity(intent);
+        ContextCompat.startActivity(mContext, intent, startAnimationBundle);
     }
 
     /**
--- a/browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchFragmentTest.java
+++ b/browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchFragmentTest.java
@@ -20,6 +20,8 @@ import org.robolectric.RobolectricTestRunner;
 
 import java.util.Collections;
 
+import androidx.core.content.ContextCompat;
+
 import static junit.framework.Assert.assertEquals;
 import static junit.framework.Assert.assertFalse;
 import static junit.framework.Assert.assertNull;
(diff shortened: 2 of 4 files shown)


quinnjn:
Thank you for the quick work! I pulled it down and ran the demo. The enter animation works great 👍. I did notice that the exit animation did not run though.

Sounds like we need to add `android.support.customtabs.extra.EXIT_ANIMATION_BUNDLE` to the bundle via this [stackoverflow post](https://stackoverflow.com/a/32379402).

quinnjn:
I was brainstorming on this some more. What do you think of overloading a 3rd `Bundle uiExtras` to `browserSwitch(Intent,String)`?

That signature is probably the most popular one since you don't need to setup Chrome Custom Tabs, or the Intent.

```
Bundle uiExtras = ActivityOptionsCompat.makeCustomAnimation(mContext, android.R.anim.fade_in, android.R.anim.fade_out).toBundle();

uiExtras.putInt("android.support.customtabs.extra.TOOLBAR_COLOR", getResources().getColor(R.color.myColor));

browserSwitch(requestCode, url, uiExtras);
```

callingmedic911:
Thanks for following up. I looked into Chrome Custom Tabs API, what I found that almost all customization/parameter are put in Intent except start activity animation, that also makes sense because animation has be known to before calling start activity.

That's why your suggested code will not work, since toolbar customization has to be set directly on intent :
```java
Bundle uiExtras = ActivityOptionsCompat.makeCustomAnimation(mContext, android.R.anim.fade_in, android.R.anim.fade_out).toBundle();
uiExtras.putInt("android.support.customtabs.extra.TOOLBAR_COLOR", getResources().getColor(R.color.myColor));
browserSwitch(requestCode, url, uiExtras);
```

Following customization should work:
```java
Intent intent = new Intent(Intent.ACTION_VIEW, returnUri)
                            .addFlags(FLAG_ACTIVITY_NEW_TASK);
            ChromeCustomTabs.addChromeCustomTabsExtras(mContext, intent);

            // Custom activity start animation
            Bundle startActivityAnimation = ActivityOptionsCompat.makeCustomAnimation(
                    mContext, android.R.anim.fade_in, android.R.anim.fade_out).toBundle();
            
            // Custom toolbar color
            intent.putExtra("android.support.customtabs.extra.TOOLBAR_COLOR",
                    getResources().getColor(android.R.color.holo_red_dark));
            
            // Custom activity exit animation
            Bundle exitActivityAnimation = ActivityOptionsCompat.makeCustomAnimation(
                    mContext, android.R.anim.fade_out, android.R.anim.fade_in).toBundle();
            intent.putExtra("android.support.customtabs.extra.EXIT_ANIMATION_BUNDLE", exitActivityAnimation);

            browserSwitch(1, intent, startActivityAnimation);
```

That's why I used browserSwitch(requestCode, intent, bundle).

Although, I agree on the fact that user should not have to handle following Custom Custom Tabs initialization work when using Intent method:
```java
Intent intent = new Intent(Intent.ACTION_VIEW, returnUri)
                            .addFlags(FLAG_ACTIVITY_NEW_TASK);
            ChromeCustomTabs.addChromeCustomTabsExtras(mContext, intent);
```
We can add this to current Intent based signature and it should be cause any backward compatibility issue even if user has pre-setup before calling. If you agree, I can put a commit on same PR.

quinnjn:
I was suggesting to add `Bundle` to the `browserSwitch(Intent, String)` signature to allow that flow to be UI customizable.

I think the changes you've made to the `browserSwitch(int, Intent, Bundle)` signature are great so far.

callingmedic911:
Are you referring to `broeserSwitch(int, String)? If not, pardon me for ignorance but I'm not able to find method with Intent, String. Can you point me to tight direction.

Thank you.

quinnjn:
Ah sorry, my bad! [`browserSwitch(int requestCode, String url)`](https://github.com/braintree/browser-switch-android/pull/16/files#diff-76675fd370f48d52965e79cdcdc68fdbR103)

callingmedic911:
For method `browserSwitch(int requestCode, Intent intent)`, I can put all customization directly on `Intent` like this:
```java
// Custom toolbar color
intent.putExtra("android.support.customtabs.extra.TOOLBAR_COLOR",
        getResources().getColor(android.R.color.holo_red_dark));
```
but for start activity animation it is required to call:
```java
ContextCompat.startActivity(context, intent, animationBundle);
```
Hence needing third argument bundle to be passed down on that method.

For method `browserSwitch(int, String)` adding `Bundle` will only enable start activity animation because user do not have access to `Intent` to add values for UI extras (toolbar customization, exit animation etc). Made no sense to only enable that, hence I left that method as it is.

For this method I'd recomend something like this:
```java
BrowserSwitchParams params = new BrowserSwitchParams();
params.setUrl("http://www.example.com");
params.setToolbarColor(R.color.customColor);
//todo set entry/exit animation and other customization
browserSwitch(params);
```
On `BrowserSwitchParams`, we will have `getUrl()`, `getIntent()` and `getBundle()` and we can use that where we require.

quinnjn:
I think the `BrowserSwitchParams` suggestion is awesome! That would be the cleanest API.

The only concern is not addressing all the customizations that are available, or available in the future. The benefit of the `Intent` is it can be passed straight through. I think it is a fair compromise to keep the API clean.

callingmedic911:
In that case, if we choose not to go with BrowserSwitchParams, I already implemented Intent and Bundle passing through.

I think you can merge this PR. Let me know if I missed anything.

quinnjn:
Sorry for the confusion but I think we do want to implement `BrowserSwitchParams`. 

I think we should create the `browserSwitch(BrowserSwitchParams)` signature and that should be the way we provide Chrome Custom Tab customizations. 

The existing `browserSwitch` signatures should behave as is, and delegate to `browserSwitch(BrowserSwitchParams)`.

Let me know if that sounds good to you!

callingmedic911:
Sounds great to me! I'll work on it and will update you.

crookedneighbor:
@callingmedic911 any update on this?

callingmedic911:
So sorry, I was not able to commit time to this. This is on our product's backlog as well, so I'll work on it soon. I'll update you as soon this is done.

PR #18, opened 2019-11-12
Use the browser switch with custom tabs in the same activity task as `BraintreeFragment`'s activity

calvarez-ov:
Motivation:

When 3DS is done in an external browser inside another activity stack, sometimes the users try to navigate back to the app using back or recents:
* Either they navigate back to the app instead of browser after reading their 3DS code (in the sms app or bank app).
* Or, perhaps more rarely, the automatic deep link redirection after a successful code entry doesn't work, and users try to manually navigate back to the app using back or recents, instead of tapping the big blue "RETURN TO APP" button.

These scenarios result in the purchase being canceled.

This PR uses custom tabs in the same activity stack as the app. When the 3DS input succeeds, and the automatic redirection works, the custom tabs are automatically closed, and the calling activity (containing `BrowserSwitchFragment`) is brought to the top of the activity stack.

If the automatic redirection doesn't succeed, we still have the possible problem of a confused user who has to tap on the "RETURN TO APP" button. But since we're in the same activity stack as the app, the user can't navigate away, so the purchase can't be canceled.

Technical details:

* Detect browsers other than Chrome which support custom tabs (this could perhaps go into a separate PR)
* Don't add these flags when launching the browser intent: `FLAG_ACTIVITY_MULTIPLE_TASK`, `FLAG_ACTIVITY_CLEAR_TASK`, `FLAG_ACTIVITY_NEW_TASK`
* Launch the browser from the activity context, not from the application context (we can't use the application context anymore as that requires `FLAG_ACTIVITY_NEW_TASK`)
* Tell `BrowserSwitchActivity` which intent to launch when it finishes (`DemoActivity` in the demo example). This is done in two places:
  - The "normal" scenario is from `BrowserSwitchFragment`, just before it launches custom tabs.
  - The other scenario is an edge case, in case the Android OS is aggressively killing our app process (reproducible by toggling "Don't keep activities" and "No background processes" in system developer settings). In this case, the application class tells `BrowserSwitchActivity` which intent to launch.
* Also to survive the app process being killed, persist the return uri to disk (shared preferences) instead of just in memory (static field). Without this, we end up with a scenario where, when `BrowserSwitchActivity` finishes itself, the app is killed, the return uri is lost, and when `BrowserSwitchFragment` resumes, it calls `onBrowserSwitchResult` with `BrowserSwitchResult.CANCELED`.

Note: I tried a few other approaches. For example, using an activity lifecycle callback listener to detect when `BrowserSwitchActivity` finishes, to relaunch `BrowserSwitchFragment`'s activity. These had various issues which can be found in the commit comments. I can squash this if it's ever ready to be merged, but I've kept the history of these attempts in git for now (mostly to be able to answer any potential questions about why other approaches may not work).


browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
```diff
@@ -140,7 +185,7 @@ public void browserSwitch(int requestCode, Intent intent) {
         }
 
         mRequestCode = requestCode;
-        mContext.startActivity(intent);
+        requireActivity().startActivity(intent);
```
calvarez-ov: Note: An alternative would be to get rid of this `mContext` field which is set to the application context. That would also require changes to the `braintree_android` repository, as `BraintreeFragment` also sets `mContext` to the application context.

I didn't know if `mContext` had to be the app context for any particular reason, so I left it this way. Because of this, I had to change the assertions in `BrowserSwitchFragmentTest` to use Robolectric apis to validate the launched intent, instead of mockito's apis on the mocked `Context`.

browser-switch/src/main/java/com/braintreepayments/browserswitch/ChromeCustomTabs.java
```diff
@@ -25,22 +36,23 @@ public static boolean isAvailable(Context context) {
         if (SDK_INT < JELLY_BEAN_MR2) {
             return false;
         }
+        // https://developer.chrome.com/multidevice/android/customtabs
```
calvarez-ov: This logic to check any compatible custom tabs browsers (not just Chrome) comes from the above documentation.

This allows custom tabs to be used in the scenario where the Samsung Internet browser is the only installed and enabled browser on the device. (Samsung Internet supports custom tabs.)

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
```diff
@@ -56,8 +58,52 @@ public void onCreate(Bundle savedInstanceState) {
...
+
+        @Override
+        public void onActivityStopped(Activity activity) { }
+
+        @Override
+        public void onActivitySaveInstanceState(Activity activity, Bundle outState) { }
+
+        @Override
+        public void onActivityDestroyed(Activity activity) {
+            // https://stackoverflow.com/questions/39726547/i-want-to-close-chrome-custom-tab-when-action-button-is-clicked
```
calvarez-ov: Note: perhaps it could make sense to have this logic (to bring the activity holding `BraintreeFragment` to the top of the stack) inside the application, instead of the browser-switch sdk. I couldn't think of a scenario where it wouldn't work the way it's implemented in this PR, but I only tested it in our app...

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
```diff
@@ -56,8 +58,52 @@ public void onCreate(Bundle savedInstanceState) {
...
+            // Relaunch ourselves, bringing ourselves to the front, when the BrowserSwitchActivity
+            // self-finishes.
+            if (activity instanceof BrowserSwitchActivity) {
+                Activity myActivity = getActivity();
+                if (myActivity != null) {
+                    Intent initialIntent = myActivity.getIntent();
+                    Intent relaunchIntent = new Intent(initialIntent)
+                            .addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+                    myActivity.startActivity(relaunchIntent);
```
calvarez-ov: Note, this requires the activity to have `android:launchMode="singleTask"` in `AndroidManifest.xml`

demerino: @calvarez-ov This sounds like a breaking change, can you clarify what would happen if `android:launchMode="singleTask"` is not included?

quinnjn:
It would be nice if Chrome gave us an option to programatically close a CCT. I've been commenting on a Chromium request to add a programatic way to close CCTs if you're interested in peeking and poking the Chromium team on that thread: https://bugs.chromium.org/p/chromium/issues/detail?id=925539#c9

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
calvarez-ov: Quick answer: without "singleTask", the custom tabs are closed, and our activity is brought to the front. But our activity doesn't detect anything about the successful payment to proceed the flow. I'll update here in a bit with more technical details/explanation. I have to investigate :)

calvarez-ov: Without "singleTask": Our activity's `onCreate()` is called, which calls `braintreeFragment = BraintreeFragment.newInstance(this, token)`. This is a new `BraintreeFragment` instance`.

With "singleTask": our activity's `onCreate()` isn't called. Looks like it's the same instance of the activity, so the same instance of the `BraintreeFragment`

calvarez-ov: Without "singleTask", our activity's `onCreate()` is called with `null` for the saved `Bundle`.  So state can't be restored. Also verified that `onNewIntent()`, and `onRestoreInstanceState()` are not called.

calvarez-ov:
Regarding the chromium issue, @quinnjn: Is that tracker specific to the implementation by the Chrome browser? Or does it also apply to the api that other browsers (like Samsung Internet) will have to implement?

quinnjn:
I tried out the demo on this branch and it doesn't look like its returning the value. 
browser-switch-actual

This is an issue with [DemoActivity](https://github.com/braintree/browser-switch-android/blob/master/demo/src/main/java/com/braintreepayments/browserswitch/demo/DemoActivity.java) since we create a DemoFragment once in onCreate:
```
@Override
    protected void onCreate(@Nullable Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        getSupportFragmentManager().beginTransaction()
                .add(android.R.id.content, new DemoFragment())
                .commit();
    }
```

but I think this demonstrates there is a behavior change in this PR where the host Activity's `onCreate` wasn't getting called with the current flow. This PR recreates the Activity and causes `onCreate` to be called again, which would be a breaking behavior change 😢

calvarez-ov:
Argh I didn’t even check the demo activity, sorry about that. I only tested
our app. Maybe it’s related to the singleTask issue discussed in previous
comments. It’s the end of the day over here and I’m not at my pc right now
so I can’t test it out right now. But maybe adding singleTask to the demo
activity may “fix” this issue. I realize it’s a breaking change which could
be problematic for apps. Maybe this behavior could be optional somehow?

demerino:
@calvarez-ov Thanks for taking the time to comment and answer questions. Maybe we can add some sort of `feature flag` like `

calvarez-ov:
I've pushed a commit for the demo activity. Indeed, setting the launch mode to `singleTask` makes the demo work again.

demo/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,8 @@
         android:icon="@mipmap/ic_launcher"
         android:label="Browser Switch Demo">
 
-        <activity android:name=".DemoActivity">
+        <activity android:name=".DemoActivity"
+            android:launchMode="singleTask">
```
quinnjn: We can't change this because it is demonstrating how a current integration is working in the wild. If we see odd behavior with DemoActivity it shows we're making a breaking change. Also changing this shows that we're adding a breaking change to the library. Ideally this activity would not change until a major version bump.

We'll need a way for consumers to opt into this new behavior flow.

calvarez-ov: That’s ok, I understand and completely agree. Feel free to use this PR for inspiration to implement a more complete solution with a feature toggle.

I'm sorry that I can't implement the full solution with the feature toggle myself. My employer says I've already spent too much time on this already 😃 and our fork is working well for us for now.

browser-switch/src/main/java/com/braintreepayments/browserswitch/ChromeCustomTabs.java
calvarez-ov: Looks like the new release 1.0.0 still has a hard restriction to only work with chrome.

calvarez-ov: Submitted #29 for this

calvarez-ov:
Replaced by:
* v1.0.0: the new version implements persisting the return url
* #29: for checking other custom tabs compatible browsers
* #30: for supporting the custom tabs in the same activity stack as `BraintreeFragment`'s activity, based on v1.0.0.

PR #19, opened 2019-11-22
Add an instrumentation test

calvarez-ov:
In case you find it useful, here's a suggestion of an instrumentation test. To catch a mistake like the one I did in PR #18 :smile: (forgetting to add `singleTask`).

----
Summary:

* Use UIAutomator instead of Espresso, since we need to interact with an external app (the browser).
* ~Add a `AndroidManifest.xml` for instrumentation tests which overrides the default `AndroidManifest.xml`, to avoid a build error regarding the `minSdkVersion`. A `minSdkVersion` of 18 is required to use UIAutomator, but this project's `minSdkVersion` is 15.~
* Use string resources, instead of string literals, in `DemoFragment`, so we can use the same strings in `DemoActivityTest`

----

* Note: I would have preferred to add the test in Kotlin, but didn't since Kotlin isn't used in this project yet. Didn't want to impose this as part of this PR.
* Note: This test also passes if integrated to the PR #18 (custom tabs in the same activity stack). It will fail in that branch if the `singleTask` mode is removed from `DemoActivity`.
* Note: ⚠️ This test relies on strings external to this repository: the strings in the [sample browser switch webpage](https://braintree.github.io/popup-bridge-example/this_launches_in_popup.html). So if anything changes in that page, this test can fail.

### DIFF
--- a/demo/build.gradle
+++ b/demo/build.gradle
@@ -29,4 +29,7 @@ dependencies {
     implementation "androidx.fragment:fragment:${rootProject.fragmentVersion}"
 
     androidTestImplementation 'com.lukekorth:device-automator:1.0.0'
+    androidTestImplementation "androidx.test:runner:$androidxTestRunnerVersion"
+    androidTestImplementation "androidx.test.ext:junit:$androidxTestJunitVersion"
+    androidTestImplementation "androidx.test.uiautomator:uiautomator:$uiAutomatorVersion"
 }
--- a/demo/src/androidTest/java/com/braintreepayments/browserswitch/demo/DemoActivityTest.java
+++ b/demo/src/androidTest/java/com/braintreepayments/browserswitch/demo/DemoActivityTest.java
@@ -1,0 +1,54 @@
+package com.braintreepayments.browserswitch.demo;
+
+import android.app.Application;
+import android.content.Context;
+import android.content.Intent;
+
+import androidx.test.core.app.ApplicationProvider;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+import androidx.test.platform.app.InstrumentationRegistry;
+import androidx.test.uiautomator.By;
+import androidx.test.uiautomator.UiDevice;
+import androidx.test.uiautomator.UiObjectNotFoundException;
+import androidx.test.uiautomator.UiSelector;
+import androidx.test.uiautomator.Until;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertTrue;
+
+
+@RunWith(AndroidJUnit4.class)
+public class DemoActivityTest {
+    private static final long APP_LAUNCH_TIMEOUT_MS = 5000;
+
+    @Test
+    public void testDemo() throws UiObjectNotFoundException {
+        UiDevice device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation());
+        Context context = ApplicationProvider.getApplicationContext();
+
+        // Launch our app using UIAutomator instead of Espresso.
+        // If we use Espresso, our activity is killed when the browser opens.
+        Intent intent = context.getPackageManager().getLaunchIntentForPackage(BuildConfig.APPLICATION_ID);
+        context.startActivity(intent);
+        device.wait(Until.hasObject(By.pkg(BuildConfig.APPLICATION_ID).depth(0)), APP_LAUNCH_TIMEOUT_MS);
+
+        performBrowserSwitch(device, "Red", "red");
+        performBrowserSwitch(device, "Green", "green");
+        performBrowserSwitch(device, "Blue", "blue");
+        performBrowserSwitch(device, "I don't like any of these colors", "undefined");
+    }
+
+    private void performBrowserSwitch(UiDevice device,
+                                      String colorActionLabel,
+                                      String expectedSelectedColor) throws UiObjectNotFoundException {
+        Application application = ApplicationProvider.getApplicationContext();
+        device.findObject(By.res(BuildConfig.APPLICATION_ID, "browser_switch")).click();
+        UiSelector selector = new UiSelector();
+        device.findObject(selector.text(colorActionLabel)).click();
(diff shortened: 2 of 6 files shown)


demo/src/androidTest/java/com/braintreepayments/browserswitch/demo/DemoActivityTest.java
```diff
@@ -0,0 +1,54 @@
...
+        performBrowserSwitch(device, "Blue", RETURN_URL_PREFIX + "?color=blue");
+        performBrowserSwitch(device, "I don't like any of these colors",
+                RETURN_URL_PREFIX + "cancel?color=undefined");
+    }
+
+    private void performBrowserSwitch(UiDevice device,
+                                      String colorActionLabel,
+                                      String expectedReturnUrl) throws UiObjectNotFoundException {
+        UiSelector selector = new UiSelector();
+        device.findObject(selector.text("Start Browser Switch")).click();
```
calvarez-ov: It probably would have been better to use a string resource id instead of `"Start Browser Switch"`. And perhaps also some string resource ids with formatting for the result and return texts. But:
* We have to use String literals anyway for interacting with the browser
* String resource ids aren't currently used in this demo, and I didn't want to change the app code just to make an instrumentation test suggestion

demo/src/androidTest/AndroidManifest.xml
```diff
@@ -0,0 +1,11 @@
...
+<!--
+https://stackoverflow.com/questions/30585289/set-different-minsdkversion-for-testandroid-than-for-main-app
+Ignore the minSdkVersion of UIAutomator (which is 18), otherwise we get a build error, due to this
+project's minSdkVersion being 15.
+-->
+<manifest xmlns:tools="http://schemas.android.com/tools"
+    package="com.braintreepayments.browserswitch.demo">
+
+    <uses-sdk tools:overrideLibrary="android_libs.ub_uiautomator" />
+</manifest>
```
calvarez-ov: An alternative to this file would be to bump `minSdkVersion` to `18`. Maybe that could actually make sense since the min sdk version of the braintree sdk is `21` anyway.

sestevens:
Hey @calvarez-ov 👋 Thanks for opening this PR! Sorry we haven't had a chance to look at it yet. We'll make sure to give it a closer review soon.

demo/src/androidTest/java/com/braintreepayments/browserswitch/demo/DemoActivityTest.java
sestevens: I'm OK with having the raw string here, but it's only working for me if this string is in all caps: `"START BROWSER SWITCH"`.

demo/src/androidTest/AndroidManifest.xml
sestevens: I'll bring this up with the team. It might be time for us to bump the min SDK version of this project.

sestevens: We think it's time for a major version bump on this repo, so we can increase `minSdkVersion` to `21`. @calvarez-ov do you want to make that change in this PR?

demo/src/androidTest/java/com/braintreepayments/browserswitch/demo/DemoActivityTest.java
calvarez-ov: I see that on some versions of Android, the OS capitalizes the button text. I fixed this to do the look up by id instead.

However, now that I started to test on different android versions, on some versions (example: marshmallow) it doesn't seem to be able to find the links like "Red". I didn't have time yet to investigate this 😞

demo/src/androidTest/AndroidManifest.xml
calvarez-ov: I think it would be better to have a separate PR for that, not tied to adding an instrumentation test.

Could your team please do that?

I think at the same time, you might update the other versions like `compileSdkVersion`, `targetSdkVersion` and `buildToolsVersion` to the latest versions.

Then I can rebase this PR on top of those changes, which will result in a smaller PR.

Thanks!

sestevens: Sure, that sounds like a good plan 👍 I will post here when we have a PR open so that you can rebase.

calvarez-ov: Didn't have news on this in a while, so I opened a separate PR #24 to increase the `minSdkVersion` to `21`.

browser-switch/build.gradle
```diff
@@ -8,7 +8,7 @@ android {
     buildToolsVersion '28.0.2'
 
     defaultConfig {
-        minSdkVersion 15
+        minSdkVersion 21
```
calvarez-ov, line 11: This change is actually part of PR #24 in branch `set-minSdkVersion-to-21`, but it appears in this PR as well, because I can't change the base branch of this PR to `set-minSdkVersion-to-21` which is only in my fork.

demo/build.gradle
```diff
@@ -6,15 +6,20 @@ android {
 
     defaultConfig {
         applicationId "com.braintreepayments.browserswitch.demo"
-        minSdkVersion 15
+        minSdkVersion 21
```
calvarez-ov, line 9: This change is actually part of PR #24 in branch `set-minSdkVersion-to-21`, but it appears in this PR as well, because I can't change the base branch of this PR to `set-minSdkVersion-to-21` which is only in my fork.

calvarez-ov:
Rebased on master.

demo/src/androidTest/AndroidManifest.xml
calvarez-ov: The minSdkVersion bump was merged in master. I've rebased this PR, so this comment outdated now.

calvarez-ov:
Rebased onto master to have the latest changes from the 1.1.0 release.

calvarez-ov:
Actually, it looks like the idea of this PR was incorporated in 11dbf187f92d0ac617cd38d3a06dcc462af5aca5, in `BrowserSwitchTest.java`

I'll close this PR then 🤷‍♀️

PR #20, opened 2020-02-03
Fix intent query and JavaScript syntax error

saschpe:
Includes a new release commit (0.2.1), which would have to be signed and published if this is accepted.
To be able to build, I had to update to a more recent Gradle Wrapper / Gradle Android plugin pair.


saschpe:
Not quite sure what to make of the test failures, they existed before...

sestevens:
Hi @saschpe, thanks for opening this PR! Could you describe the issue this fixes and give us steps to reproduce it?

saschpe:
Sure, these screenshots display the difference with or without using CATEGORY_DEFAULT:

Screenshot 2020-02-03 at 09 10 55
Screenshot 2020-02-03 at 09 10 41

I would have to grab another screenshot via Chrome DevTools about the error message issue. Basically, when there is no intent found, you raise an error with that message. That gets evaluated in a browser environment which replaces double quotes with single quotes (nothing that can be influenced). Thus, single quotes can't be part of the string.

saschpe:
Any update on this?

sshropshire:
Hi @saschpe thanks for using the Braintree SDK.

According to [Google](https://developer.android.com/training/app-links/deep-linking#adding-filters), `Intent.CATEGORY_DEFAULT` is required to allow an app to respond to implicit intents, i.e. a re-direct from the browser.

Is there a particular use case in your app where this breaks functionality? We have a recommended setup in our [docs](https://developer.android.com/training/app-links/deep-linking#adding-filters) that enables successful browser switching through our SDK.

saschpe:
The documentation you are quoting is for the target activity to declare in it's AndroidManifest.xml. On the querying side, only ACTION_VIEW plus URL matters. One could axe CATEGORY_BROWSEABLE as well but it doesn't seem to hurt. 

See also [PackageManager](https://developer.android.com/reference/android/content/pm/PackageManager#queryIntentActivities(android.content.Intent,%20int)) and the the [Intents training](https://developer.android.com/training/basics/intents/sending#StartActivity).

sshropshire:
@saschpe this fix breaks functionality for our use case. The query is an assertion to ensure that the browser switch target Activity is properly configured in the manifest. It's a safe guard to ensure that browser switch is properly configured.

If you remove `Intent.CATEGORY_DEFAULT` in your manifest you can observe browser switch will stop working. Our SDK aims to catch this misconfiguration and notify the developer with a runtime error.

saschpe:
> Our SDK requires line 164 of BrowserSwitchFragment.java to be present. We can merge once this fix has been added.

Can you elaborate a bit more why this is the case? Both the provided **demo** app in this repo and  **PopupBridgeExample** in _popup-bridge-android_ continue to work just fine with this change. Without, an arguably more complex app fails. Unfortunately, it's not public so I can't share much beyond the screenshots above. I could go to greater lengths and try to extract a demo project.

BTW. I separated the undisputed changes into another PR. Those could already be merged and reduce the delta of this one until we come to a conclusion ;-)

sshropshire:
@saschpe Correct, it will work because you have properly configured your manifest to support CATEGORY_DEFAULT. If you remove this from your Browser Switch Activity's manifest, it will stop working.

Our assertion catches this use case, and provides an error message when an Activity is missing configuration to support Browser Switch. If we take CATEGORY_DEFAULT out of the query, it will fail silently.

sestevens:
@saschpe Steps to reproduce the issue you're seeing would be really helpful! From our end, everything seems to be working when we include `Intent.CATEGORY_DEFAULT`, but it seems like you've found a case where things don't work.

PR #22, opened 2020-04-16
Extract BrowserSwitchClient from BrowserSwitchFragment

sshropshire:
### Summary of changes

 - Extract BrowserSwitchClient from BrowserSwitchFragment
 - Configure BrowserSwitchFragment to use the new client

 ### Checklist

 - [x] Added a changelog entry

- @joshlarson @sshropshire


browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchAdapter.java
```diff
@@ -0,0 +1,101 @@
...
+                BrowserSwitchClient.newInstance(), BrowserSwitchConfig.newInstance());
+    }
+
+    @VisibleForTesting
+    static BrowserSwitchAdapter newInstance(BrowserSwitchClient simpleClient, BrowserSwitchConfig config) {
+        return new BrowserSwitchAdapter(simpleClient, config);
+    }
+
+    private BrowserSwitchConfig config;
+    private BrowserSwitchClient simpleClient;
```
scannillo, line 24: Why the name `simpleClient`? Just curious if this has a greater meaning?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    void start(int requestCode, Intent intent, FragmentActivity activity, BrowserSwitchListener listener) {
+        Context applicationContext = activity.getApplicationContext();
+        String errorMessage = assertCanPerformBrowserSwitch(requestCode, applicationContext, intent);
+
+        if (errorMessage != null) {
+            BrowserSwitchResult result =
+                new BrowserSwitchResult(BrowserSwitchResult.STATUS_ERROR, errorMessage);
+            listener.onBrowserSwitchResult(requestCode, result,null);
```
scannillo, line 37:
```suggestion
                listener.onBrowserSwitchResult(requestCode, result,null);
```

scannillo, line 37:
```suggestion
            listener.onBrowserSwitchResult(requestCode, result, null);
```

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchResult.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public int getStatus() {
+        return status;
+    }
+
+    public String getErrorMessage() {
+        return errorMessage;
+    }
+
+    private int status;
```
scannillo, line 16: I notice you've been putting private properties below public methods. Across the other Android SDKs we have both public and private properties at the top of the file. Are we looking to move away from this pattern?

sshropshire: @scannillo Good call, that's an oversight. Right now looking to update BraintreeFragmentUnitTest to fit the new client model and then will do a second pass on formatting.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchPersistentStore.java
```diff
@@ -0,0 +1,42 @@
+package com.braintreepayments.browserswitch;
+
+import android.content.Context;
+
+import org.json.JSONException;
+
+public class BrowserSwitchPersistentStore {
```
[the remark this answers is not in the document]
sshropshire: Yeah technically there should only be one request at a time. The BrowserSwitchClient enforces this externally by calling `getActiveRequest` and `putActiveRequest`. The idea was to match the previous implementation and help ease the strain of migration.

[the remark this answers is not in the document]
sshropshire: Here: https://github.com/braintree/browser-switch-android/blob/63e709ac090ff39097798ed3bccf9da345a180c3/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java#L104

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchAdapter.java
sshropshire: Yeah this is legacy. Throwback to a time where there wasn't a better name. About to fix.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private boolean canOpenUrl(Context context, Intent intent) {
+        return activityFinder.canResolveActivityForIntent(context, intent);
+    }
+
+    public BrowserSwitchRequest getActiveRequest(Context context) {
+        return persistentStore.getActiveRequest(context);
+    }
+
+    public BrowserSwitchRequest takeActiveRequest(Context context) {
```
scannillo, line 83: Why is this function needed? How is it different than `clearActiveRequest()`?

Also - the name `take` maybe is confusing

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchClientTest.java
```diff
@@ -0,0 +1,309 @@
...
+        request = mock(BrowserSwitchRequest.class);
+
+        // global scaffold
+        when(activity.getApplicationContext()).thenReturn(applicationContext);
+
+        sut = BrowserSwitchClient.newInstance(browserSwitchConfig, activityFinder, persistentStore);
+    }
+
+    @Test
+    public void start_whenRequestCodeIsIntegerMinValue_callsbackListenerWithError() {
```
scannillo, line 71:
```suggestion
    public void start_whenRequestCodeIsInNegativeInteger_callsbackListenerWithError() {
```

sshropshire: This one's interesting the OG implementation used `Integer.MIN_VALUE` almost like a special distinct value. I agree we probably will never have a negative request code, but it is legit checking that request code equals the minimum possible integer that can be represented on the device.

scripts/start-local-development.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
```
scannillo, line 2: What is the use of this file?

demo/src/main/res/values/strings.xml
```diff
@@ -0,0 +1 @@
+<resources></resources>
```
scannillo, line 1: Erroneous file addition?

scannillo:
Will we update this repo's demo app to include use `BrowserSwitchClient`?

scripts/start-local-development.sh
sshropshire: It polls the code base for changes, re-builds, and deploys to local maven continuously using Gradle. You can hop back and forth to the core Android project more easily and make changes (still have to press "Gradle Sync" in the core project though).

demo/src/main/res/values/strings.xml
sshropshire: Yeah this one not sure but it was probably generated. Deleted.

sshropshire:
@scannillo that's kind of where I'm on the fence. I think it makes sense, but one of the main goals was to make the refactor backward compatible.

On the other hand, once the BraintreeClient PR gets merged in, there won't be a need for BrowserSwitchFragment any more, unless we decide there's a need for the new BrowserSwitch to be backward compatible with older Core SDK versions.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchPersistentStore.java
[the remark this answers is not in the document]
sshropshire: Also yeah this class was added because it made testing really easy since its only concern is to persist BrowserSwitch requests so they can be restored when the app returns to the foreground.

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchClientTest.java
scannillo: Ah, ok - yeh I see now in the old code how we were setting the request code to `Integer.MIN_VALUE` in cases where there was a BrowserSwitch error or while BrowserSwitch was occurring.

scannillo: So this test is more of a "backwards compatibility" test.

scannillo:
> @scannillo that's kind of where I'm on the fence. I think it makes sense, but one of the main goals was to make the refactor backward compatible.
> 
> On the other hand, once the BraintreeClient PR gets merged in, there won't be a need for BrowserSwitchFragment any more, unless we decide there's a need for the new BrowserSwitch to be backward compatible with older Core SDK versions.

Sorry, I deleted my original comment which was "Will we add BrowserSwitchClient to the demo app". I deleted it, because I figured showcasing the backwards compatibility was the reason. 

But sooner than later we'll want a new demo utilizing `BrowserSwitchClient`.

Also - I agree, I don't think there is a need for new BrowserSwitch to be backward compatible with older Core SDKs.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchAdapter.java
```diff
@@ -0,0 +1,109 @@
...
+    @VisibleForTesting
+    static BrowserSwitchAdapter newInstance(BrowserSwitchClient internalClient, BrowserSwitchConfig config) {
+        return new BrowserSwitchAdapter(internalClient, config);
+    }
+
+    private BrowserSwitchConfig config;
+    private BrowserSwitchClient internalClient;
+
+    private BrowserSwitchAdapter(BrowserSwitchClient internalClient, BrowserSwitchConfig config) {
+        this.internalClient = internalClient;
```
scannillo, line 27: What was the motivation for creating `BrowserSwitchAdapter` instead of just `BrowserSwitchClient`?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchAdapter.java
```diff
@@ -0,0 +1,109 @@
...
+import android.content.Intent;
+import android.net.Uri;
+
+import androidx.annotation.VisibleForTesting;
+import androidx.fragment.app.Fragment;
+import androidx.fragment.app.FragmentActivity;
+
+public class BrowserSwitchAdapter implements IBrowserSwitchClient {
+
+    static BrowserSwitchAdapter newInstance() {
```
scannillo, line 13: Once we do get rid of `BrowserSwitchFragment`, is the plan to also get rid of `BrowserSwitchAdapter` and consolidate to `BrowserSwitchClient`?

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchClientTest.java
sshropshire: Yeah we may revise going forward. Trying to make the new code line up with the previous implementation will help with the review though I'm thinking.

The parts that need changing should stick out more than the rest as we move to refactor the other modules we can prioritize the behavioral changes.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchAdapter.java
sshropshire: We can technically but I think the adapter pattern works well here because it allows BrowserSwitchClient to only have to consider one method internally. The adapter here allows us to add additional overloads more easily as the API evolves along with the Android SDK.

sshropshire: This was also another component that made the test for `start()` a bit easier in [BrowserSwitchClient](https://github.com/braintree/browser-switch-android/blob/0618df674f1aee42823408032dd5f32a0e09636f/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java#L30).

We might be able to achieve something similar to iOS in the future by having an `IBraintreeClientInternal` interface that extends from `IBraintreeClient` and draws a clean line between internal methods/properties and what we want visible to merchants.

browser-switch/build.gradle
```diff
@@ -11,17 +17,55 @@ android {
...
+    testImplementation 'org.powermock:powermock-module-junit4-rule:2.0.7'
+    testImplementation 'org.powermock:powermock-classloading-xstream:2.0.7'
+    testImplementation 'org.powermock:powermock-api-mockito2:2.0.7'
+
+    testImplementation 'org.robolectric:robolectric:4.3'
+
+    def fragment_version = "1.2.4"
+    debugImplementation "androidx.fragment:fragment-testing:${fragment_version}"
+
+    // Ref: https://developer.android.com/training/testing/unit-testing/instrumented-unit-tests
```
sestevens, line 64: Do we need all the references in this file? It's nice to link out to references when the code is doing something complicated or confusing, but it doesn't seem like that's necessary here.

sshropshire: Yeah technically no. I usually include refs if I feel like a specific piece of code / config will induce a "WTF" and I add the links for closure lol.

browser-switch/src/main/java/com/braintreepayments/browserswitch/ActivityFinder.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public static ActivityFinder newInstance() {
+        return new ActivityFinder();
+    }
+
+    private ActivityFinder() {}
+
+    public boolean canResolveActivityForIntent(Context context, Intent intent) {
+        int numActivities = context.getPackageManager().queryIntentActivities(intent, 0).size();
+        return (numActivities > 0);
```
sestevens, line 16: Small suggestion: 
```suggestion
        return !context.getPackageManager().queryIntentActivities(intent, 0).isEmpty();
```

browser-switch/build.gradle
sestevens: Yeah, that makes sense. So what's the "WTF" here?

sshropshire: Gradle lol. I'll take 'em out.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchAdapter.java
scannillo: Gotcha, thanks for explaining!

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
```diff
@@ -1,44 +1,27 @@
...
+import androidx.annotation.VisibleForTesting;
+import androidx.appcompat.app.AppCompatActivity;
 
 /**
  * <a href="https://developer.android.com/guide/topics/manifest/activity-element.html#lmode">singleTask</a>
  * Activity used to receive the response from a browser switch. This Activity contains no UI and
  * finishes during {@link Activity#onCreate(Bundle)}.
  */
-public class BrowserSwitchActivity extends Activity {
+public class BrowserSwitchActivity extends AppCompatActivity {
```
sestevens, line 14: - How come this class is extending from `AppCompatActivity` now?
- Is this a breaking change?
- Will this cause any issues for merchants who want to use `FragmentActivity`?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
```diff
@@ -1,44 +1,27 @@
...
-     */
-    @Nullable
-    public static Uri getReturnUri() {
-        return sReturnUri;
-    }
-
-    /**
-     * Clears the return uri.
-     */
-    public static void clearReturnUri() {
```
sestevens, line 41: I see we're removing a couple public methods here - isn't this a breaking change?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
sshropshire: Hey yeah, good point I think this can be reverted. AppCompatActivity is the recommended approach now, but BrowserSwitchActivity is headless, there's no associated UI. It mainly exists to capture the deep link from the browser when re-directing back into the app.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
sshropshire: Yeah these are public but the merchant shouldn't ever have to use BrowserSwitchActivity directly. Also, this logic now lives in BrowserSwitchClient.

Going forward, we may allow breaking changes for this next version.

browser-switch/build.gradle
```diff
@@ -46,6 +83,9 @@ artifacts {
 }
 
 signing {
+    required {
```
scannillo, line 86: This needs to be removed before release, I assume?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
```diff
@@ -0,0 +1,332 @@
...
+            BrowserSwitchConfig.newInstance(), ActivityFinder.newInstance(),
+            BrowserSwitchPersistentStore.getInstance(), returnUrlScheme
+        );
+    }
+
+    @VisibleForTesting
+    static BrowserSwitchClient newInstance(
+        BrowserSwitchConfig config, ActivityFinder activityFinder,
+        BrowserSwitchPersistentStore persistentStore, String returnUrlScheme
+    ) {
```
scannillo, line 31: Having the `{` on a newline isn't consistent with rest of repo

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
```diff
@@ -0,0 +1,332 @@
...
+
+    /**
+     * Open a browser or <a href="https://developer.chrome.com/multidevice/android/customtabs">Chrome Custom Tab</a>
+     * with the given url from an Android activity.
+     *
+     * @param requestCode the request code used to differentiate requests from one another.
+     * @param uri the url to open.
+     * @param activity the activity used to start browser switch
+     */
+    @SuppressWarnings("SameParameterValue")
```
scannillo, line 87: Why was this needed?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
```diff
@@ -0,0 +1,332 @@
...
+     * Deliver a pending browser switch result to an Android fragment that is also a BrowserSwitchListener.
+     * We recommend you call this method in onResume to receive a browser switch result once your
+     * app has re-entered the foreground.
+     *
+     * Cancel and Success results will be delivered only once. If there are no pending
+     * browser switch results, this method does nothing.
+     *
+     * @param fragment the BrowserSwitchListener that will receive a pending browser switch result
+     */
+    void deliverResult(Fragment fragment) {
```
scannillo, line 240: Should these `deliverResult()`s all be marked as public?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
scannillo: If we're already introducing breaking change by removing public methods like this - why do we need to keep BrowserSwitchActivity and BrowserSwitchFragment around at all anymore? I am a bit confused on how how some of this PR will be backwards compatible and some wont.

scannillo:
Can we include in this PR a `REAME.md` update for the new way of suggested integration?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
sshropshire: Yeah the main backward compatible portion is BrowserSwitchFragment. That's the only part that the merchant interacts with (transitively, via BraintreeFragment).

BrowserSwitchActivity has public access, but it's technically not merchant facing. We don't have anything in the docs about BrowserSwitchActivity except for adding it to the AndroidManifest.xml to allow deep links to work.

browser-switch/build.gradle
sshropshire: I think it's fine. This snippet came from braintree_android originally.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
sshropshire: Yeah this is something to address in our style guidelines. I'm up for placing the `) {` as a trailing charset, the main deal here was to try to meet the 100 column limit.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
sshropshire: Android Studio flagged a warning because the demo project only calls this method once. The IDE thinks the parameter `requestCode` could potentially be a constant. I suppressed the warning to make the code lint-free, but we may be able to ignore this linting error with a lint config in the future.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
sshropshire: We can. I was marking methods public on an as-needed basis.

sshropshire:
@scannillo thanks that makes a lot of sense. I just added sample code to the README.

build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.4'
+        classpath 'com.android.tools.build:gradle:3.6.3'
```
hollabaq86, line 7: should this version bump be noted in the CHANGELOG?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
sshropshire: Yeah I agree as well. I'm up for the bump. Removing these methods cleans up the code a bit since we're no longer relying on global state for Browser Switch to work. I think that warrants a breaking change in this case and gets us to v1.0.

.travis.yml
```diff
@@ -9,6 +9,7 @@ android:
     - tools
     - tools
     - platform-tools
-    - build-tools-28.0.2
+    - build-tools-28.0.3
     - android-28
 script: ./gradlew --info clean lint test
+
```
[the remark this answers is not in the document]
sshropshire: I'm up for establishing a convention to remove trailing new lines on files.

Wanted to point out this is an OG Unix thing: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchResult.java
```diff
@@ -0,0 +1,28 @@
+package com.braintreepayments.browserswitch;
+
+public class BrowserSwitchResult {
+
+    final private int status;
+    final private String errorMessage;
+
+    public static final int STATUS_OK = 1;
+    public static final int STATUS_CANCELLED = 2;
```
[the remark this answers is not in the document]
sshropshire: 👌

build.gradle
sshropshire: Yeah definitely

README.md
```diff
@@ -87,6 +87,44 @@ public void onBrowserSwitchResult(int requestCode, BrowserSwitchResult result, @
 }
 ```
 
+## BrowserSwitchClient
```
scannillo, line 90: This header is at the same level as `##Usage`. Should we make these two integration patterns their own unique headers? Also - I assume we prefer merchants to utilize the new pattern, so maybe we put it first.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
scannillo: @sshropshire Did we come to a conclusion on this? Holly makes a good point that devs use this SDK who aren't necessarily using Braintree. Even though these changes don't affect BT merchants, they're potential breaking changes to other devs using this SDK. Can we note these updates in the CHANGELOG as well?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchPersistentStore.java
```diff
@@ -0,0 +1,38 @@
...
+    static BrowserSwitchPersistentStore getInstance() {
+        return INSTANCE;
+    }
+
+    private BrowserSwitchPersistentStore() {}
+
+    BrowserSwitchRequest getActiveRequest(Context context) {
+        BrowserSwitchRequest request = null;
+
+        String activeRequestJson = PersistentStore.get("browserSwitch", context);
```
billwerges, line 20: `browserSwitch` seems like a good candidate for a constant in this class. It could also be used to replace all the occurrences of this string within the tests.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchResult.java
```diff
@@ -0,0 +1,28 @@
+package com.braintreepayments.browserswitch;
+
+public class BrowserSwitchResult {
+
+    final private int status;
+    final private String errorMessage;
+
+    public static final int STATUS_OK = 1;
+    public static final int STATUS_CANCELED = 2;
+    public static final int STATUS_ERROR = 3;
```
billwerges, line 10: Would it make sense to wrap these 3 values in an `enum` type?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
sshropshire: Yeah, I realize now that we would need to add Breaking Changes to the CHANGELOG I'll make sure to update those. I think we decided to just go for the breaking change here and make it a major version bump deal.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchPersistentStore.java
sshropshire: Yeah definitely will create a constant and update the class.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchResult.java
sshropshire: They were enums originally I believe, but Enums on Android kind of have a [bad rap](https://android.jlelse.eu/android-performance-avoid-using-enum-on-android-326be0794dc3). During the refactoring seemed like a good opportunity to remove them.

README.md
sshropshire: Yeah this is a good question. It's kind of an alternate usage right now. The original integration is probably the most seamless at the moment, but we need more control over how BrowserSwitchListener gets injected right now.

I think in v1.1.0, we may be able to leverage Activity Lifecycles to get a similarly seamless interface without having to have access to BrowserSwitchFragment. For now maybe `Alternative Usage: BrowserSwitchClient` would be a good title?

scannillo: Sounds good

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchResult.java
billwerges: I'm good with not using an `enum` if it negatively impacts performance. My main concern was the lack of value checking since the constructors accept plain `int` values. Have we considered using the Android `Typedef` annotations as an `enum` replacement?

https://developer.android.com/studio/write/annotations#enum-annotations

sshropshire: Yeah, this is a legitimate concern. I tried the enum annotations out once before and ran into some weirdness with mockito. I'll give it another shot to see if I can get them to work.

PR #24, opened 2020-04-17
Increase minSdkVersion from 15 to 21

calvarez-ov:


sshropshire:
Hello @calvarez-ov thank you for your PR. We had another PR that bumps the minSdkVersion as well, so all should be good in master. We will be releasing the new version soon.

PR #25, opened 2020-06-22
Create Rakefile

sshropshire:
### Summary of changes

 - Add Rakefile to project
 - Delete `release.sh` script
 - Add Issue and PR templates
 
 ### Checklist

 ~- [ ] Added a changelog entry~

- @scannillo @sshropshire


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+Thank you for your contribution to Braintree. 
+
+> Before submitting this PR, note we cannot accept language translation PRs. We support the same languages that are supported by PayPal, and have a dedicated localization team to provide the translations. If there is an error in a specific translation, you may open an issue and we will escalate it to the localization team.
```
sestevens, line 3: Just checking - are translations relevant to browser switch? I can't remember if there are any user-facing strings. If not, we can probably remove this note.

sshropshire: Yeah, we were on the fence about leaving this but it's probably best to just remove.

sestevens: Cool cool. Yeah, we can always add it back in if we need to.

PR #26, opened 2020-06-23
Maven Publish

sshropshire:
### Summary of changes

 - Upgrade Android Gradle Plugin to version 4.0.0
 - Upgrade Gradle to version 6.1.1
 - Update Android Build Tools to version 29.0.2
 - Migrate to `maven-publish` plugin
 - Update Rakefile release to use updated `publishToSonatype` task

 ### Checklist

 - [x] Added a changelog entry

- @scannillo @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,5 @@
 # browser-switch-android Release Notes
 
-## unreleased
-
-* Upgrade Android Gradle Plugin to version 4.0.0
-* Upgrade Gradle to version 6.1.1
-* Update Android Build Tools to version 29.0.2
-
 ## 1.0.0
 
 * Create BrowserSwitchClient to allow browser switch behavior through composition as well as inheritance.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # browser-switch-android Release Notes
 
+## unreleased
+
+* Upgrade Android Gradle Plugin to version 4.0.0
+* Upgrade Gradle to version 6.1.1
+* Update Android Build Tools to version 29.0.2
+
 ## 1.0.0
 
 * Create BrowserSwitchClient to allow browser switch behavior through composition as well as inheritance.
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2020 Braintree, a division of PayPal, Inc.
+Copyright (c) 2017-2020 Braintree, a division of PayPal, Inc.
 
 Permission is hereby granted, free of charge, to any person
 obtaining a copy of this software and associated documentation
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2017 Braintree, a division of PayPal, Inc.
+Copyright (c) 2020 Braintree, a division of PayPal, Inc.
 
 Permission is hereby granted, free of charge, to any person
 obtaining a copy of this software and associated documentation
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2017-2020 Braintree, a division of PayPal, Inc.
+Copyright (c) 2017 Braintree, a division of PayPal, Inc.
 
 Permission is hereby granted, free of charge, to any person
 obtaining a copy of this software and associated documentation
--- a/browser-switch/build.gradle
+++ b/browser-switch/build.gradle
@@ -90,7 +90,7 @@ nexusStaging {
 
 signing {
     required {
-        !version.endsWith("DEVELOPMENT")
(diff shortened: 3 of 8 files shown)


browser-switch/build.gradle
```diff
@@ -77,76 +78,73 @@ artifacts {
...
+                        }
                     }
-                }
-            }
 
-            pom.whenConfigured { MavenPom pom ->
-                pom.dependencies.each {
-                    if (it.artifactId == 'support-annotations') {
-                        it.optional = true
+                    licenses {
```
demerino, line 129: Maybe its a good time to update the dates in https://github.com/braintree/browser-switch-android/blob/master/LICENSE

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # browser-switch-android Release Notes
 
+## unreleased
+
+* Upgrade Android Gradle Plugin to version 4.0.0
+* Upgrade Gradle to version 6.1.1
```
demerino, line 6: Does this actually impact the merchant integration and need to be included in the CHANGELOG? Or is it just used for the build/distribution side of things?

browser-switch/build.gradle
```diff
@@ -77,76 +78,73 @@ artifacts {
...
-        !version.endsWith("DEVELOPMENT")
-    }
-    sign configurations.archives
-}
+def sonatypeUsername = System.env['SONATYPE_USERNAME']
+def sonatypePassword = System.env['SONATYPE_PASSWORD']
 
 nexusStaging {
     packageGroup = 'com.braintreepayments'
+    stagingProfileId = 'd2424106a9f1f'
```
demerino, line 86: Where did this come from?

browser-switch/build.gradle
```diff
@@ -77,76 +78,73 @@ artifacts {
...
+                artifactId = 'browser-switch'
                 version = "${android.defaultConfig.versionName}"
-                packaging 'aar'
-                description 'A library making it easy to open a url in a browser or Chrome Custom Tab and receive a response from the web page.'
 
-                url 'https://github.com/braintree/browser-switch-android'
+                pom {
+                    name = 'browser-switch'
+                    packaging = 'aar'
+                    description = 'A library making it easy to open a url in a browser or Chrome Custom Tab and receive a response from the web page.'
```
demerino, line 113: Do we want to align this with the README?

> Android Browser Switch makes it easy to open a url in a browser or Chrome Custom Tab and receive a response as the result of user interaction, either cancel or response data from the web page.

CHANGELOG.md
sshropshire: We were thinking it could affect a merchant's CI somewhere, possibly. We added it in for good measure but technically it may be omitted. A good point here because we will probably replicate this convention across the other repos as well.

I'm up for removing these since the Android Gradle Plugin and Gradle version are just dependencies. It would also make CHANGELOG easier to maintain.

browser-switch/build.gradle
sshropshire: We had to run `./gradlew getStagingProfile` from the root project. There's supposed to be a way where the plugins get the stagingProfileId dynamically, but setting it statically here helped us get it to work.

We should be able to refine the gradle file a bit more in the future.

browser-switch/build.gradle
sshropshire: Def.

browser-switch/build.gradle
sshropshire: Yeah this is also a good call.

browser-switch/build.gradle
scannillo: > The Profile ID displays the unique identifier that can be used for staging to this repository using the Nexus Staging Maven plugin. 

>  Explicit or Implicit is the default behavior and causes the repository manager to select the profile by the provided staging profile identifier and to fall back to an automatic determination, if none is provided. It is necessary to be used with the Maven deploy plugin and the correct staging profile is determined using repository targets together with the generic deploy URL.

([docs reference](https://help.sonatype.com/repomanager2/staging-releases/configuring-the-staging-suite#ConfiguringtheStagingSuite-Overview))

In short, we weren't able to successfully deploy without it. Though the reasoning for requiring it is still fuzzy to me.

CHANGELOG.md
scannillo: We will remove these CHANGELOG entries entirely

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2017 Braintree, a division of PayPal, Inc.
+Copyright (c) 2017-2020 Braintree, a division of PayPal, Inc.
```
crookedneighbor, line 1: Does this need to be updated? I don't think you need to include the current year here.

sshropshire:
```suggestion
Copyright (c) 2017 Braintree, a division of PayPal, Inc.
```

PR #27, opened 2020-07-27
Request Metadata via BrowserSwitchOptions

sshropshire:
### Summary of changes

 - Create `BrowserSwitchOptions` value object for configuring browser switch behavior
 - Add BrowserSwitchClient#start overloads with `BrowserSwitchOptions` param
 - Add BrowserSwitchFragment#browserSwitch overload with `BrowserSwitchOptions` param

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop @sshropshire

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
@@ -9,8 +9,6 @@ import androidx.annotation.VisibleForTesting;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
 
-import org.json.JSONObject;
-
 @SuppressWarnings("WeakerAccess")
 public class BrowserSwitchClient {
 
@@ -50,16 +48,11 @@ public class BrowserSwitchClient {
      * @param fragment the fragment used to start browser switch
      */
     public void start(int requestCode, Uri uri, Fragment fragment) {
-        if (fragment instanceof BrowserSwitchListener) {
-            start(requestCode, uri, fragment, (BrowserSwitchListener) fragment);
-        } else {
-            throw new IllegalArgumentException("Fragment must implement BrowserSwitchListener.");
-        }
-    }
+        BrowserSwitchOptions browserSwitchOptions = new BrowserSwitchOptions()
+                .requestCode(requestCode)
+                .url(uri);
 
-    // TODO: write docs
-    public void start(BrowserSwitchOptions browserSwitchOptions, Fragment fragment) {
-        // TODO: implement
+        start(browserSwitchOptions, fragment);
     }
 
     /**
@@ -72,17 +65,11 @@ public class BrowserSwitchClient {
      * @param listener the listener that will receive browser switch callbacks
      */
     public void start(int requestCode, Uri uri, Fragment fragment, BrowserSwitchListener listener) {
-        FragmentActivity activity = fragment.getActivity();
-        if (activity != null) {
-            start(requestCode, uri, activity, listener);
-        } else {
-            throw new IllegalStateException("Fragment must be attached to an activity.");
-        }
-    }
+        BrowserSwitchOptions browserSwitchOptions = new BrowserSwitchOptions()
+                .requestCode(requestCode)
+                .url(uri);
 
-    // TODO: write docs
-    public void start(int requestCode, Uri uri, Fragment fragment, BrowserSwitchListener listener, JSONObject metadata) {
-        // TODO: implement
+        start(browserSwitchOptions, fragment, listener);
     }
 
     /**
@@ -95,16 +82,11 @@ public class BrowserSwitchClient {
      */
     @SuppressWarnings("SameParameterValue")
     public void start(int requestCode, Uri uri, FragmentActivity activity) {
-        if (activity instanceof BrowserSwitchListener) {
-            start(requestCode, uri, activity, (BrowserSwitchListener) activity);
-        } else {
-            throw new IllegalArgumentException("Activity must implement BrowserSwitchListener.");
-        }
-    }
+        BrowserSwitchOptions browserSwitchOptions = new BrowserSwitchOptions()
+                .requestCode(requestCode)
+                .url(uri);
 
-    // TODO: write docs
-    public void start(int requestCode, Uri uri, FragmentActivity activity, JSONObject metadata) {
-        // TODO: implement
+        start(browserSwitchOptions, activity);
     }
 
     /**
@@ -116,17 +98,12 @@ public class BrowserSwitchClient {
      * @param activity the activity used to start browser switch
      * @param listener the listener that will receive browser switch callbacks
      */
-    public void start(
-            int requestCode, Uri uri, FragmentActivity activity,BrowserSwitchListener listener) {
-        Context appContext = activity.getApplicationContext();
-        Intent intent = config.createIntentToLaunchUriInBrowser(appContext, uri);
-        start(requestCode, intent, activity, listener);
-    }
+    public void start(int requestCode, Uri uri, FragmentActivity activity, BrowserSwitchListener listener) {
+        BrowserSwitchOptions browserSwitchOptions = new BrowserSwitchOptions()
+                .requestCode(requestCode)
+                .url(uri);
 
-    // TODO: write docs
-    public void start(int requestCode, Uri uri, FragmentActivity activity,
-                      BrowserSwitchListener listener, JSONObject metadata) {
-        // TODO: implement
+        start(browserSwitchOptions, activity, listener);
     }
 
     /**
@@ -138,16 +115,11 @@ public class BrowserSwitchClient {
      * @param fragment the fragment used to start browser switch
      */
     public void start(int requestCode, Intent intent, Fragment fragment) {
-        if (fragment instanceof BrowserSwitchListener) {
-            start(requestCode, intent, fragment, (BrowserSwitchListener) fragment);
-        } else {
-            throw new IllegalArgumentException("Fragment must implement BrowserSwitchListener.");
-        }
-    }
+        BrowserSwitchOptions browserSwitchOptions = new BrowserSwitchOptions()
+                .intent(intent)
+                .requestCode(requestCode);
 
-    // TODO: write docs
-    public void start(int requestCode, Intent intent, Fragment fragment, JSONObject metadata) {
-        // TODO: implement
+        start(browserSwitchOptions, fragment);
     }
 
     /**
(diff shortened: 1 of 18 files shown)


browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
```diff
@@ -112,73 +113,145 @@ public void start(
...
+            JSONObject metadata = browserSwitchOptions.getMetadata();
             BrowserSwitchRequest request = new BrowserSwitchRequest(
-                    requestCode, intent.getData(), BrowserSwitchRequest.PENDING);
+                    requestCode, intent.getData(), BrowserSwitchRequest.PENDING, metadata);
             persistentStore.putActiveRequest(request, appContext);
             appContext.startActivity(intent);
         } else {
             BrowserSwitchResult result =
-                new BrowserSwitchResult(BrowserSwitchResult.STATUS_ERROR, errorMessage);
+                    new BrowserSwitchResult(BrowserSwitchResult.STATUS_ERROR, errorMessage);
```
scannillo, line 254: Extra whitespace? Or is this intentional?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
```diff
@@ -65,7 +68,15 @@ public void browserSwitch(int requestCode, String url) {
...
     @SuppressWarnings("WeakerAccess")
     public void browserSwitch(int requestCode, Intent intent) {
-        browserSwitchClient.start(requestCode, intent, this);
+        BrowserSwitchOptions browserSwitchOptions = new BrowserSwitchOptions()
+                .intent(intent)
+                .requestCode(requestCode);
+        browserSwitchClient.start(browserSwitchOptions, this);
+    }
+
+    //TODO: write docs
```
scannillo, line 77: TODO

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchOptions.java
```diff
@@ -0,0 +1,50 @@
+package com.braintreepayments.browserswitch;
+
+import android.content.Intent;
+import android.net.Uri;
+
+import org.json.JSONObject;
+
+public class BrowserSwitchOptions {
```
scannillo, line 8: Need docstrings on this entire class

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchResult.java
```diff
@@ -18,13 +21,14 @@
     public static final int STATUS_CANCELED = 2;
     public static final int STATUS_ERROR = 3;
 
-    BrowserSwitchResult(@BrowserSwitchStatus int status) {
-        this(status, null);
+    BrowserSwitchResult(@BrowserSwitchStatus int status, String errorMessage) {
```
scannillo, line 24: Why do we need this new `errorMessage` param here? I see we're passing in `null` for it in the `deliverResult()` function above.

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchClientTest.java
```diff
@@ -304,41 +514,40 @@ public void deliverResult_whenRequestIsPending_clearsResultStoreAndNotifiesResul
...
 
     @Test
     public void deliverResult_whenFragmentIsNotABrowserSwitchListener_throwsIllegalArgumentException() {
         exceptionRule.expect(IllegalArgumentException.class);
         exceptionRule.expectMessage("Fragment must implement BrowserSwitchListener.");
 
-        Fragment fragment = mock(Fragment.class);
         sut = BrowserSwitchClient.newInstance(browserSwitchConfig, activityFinder, persistentStore, returnUrlScheme);
-        sut.deliverResult(fragment);
+        sut.deliverResult(plainFragment);
```
scannillo, line 540: What is the meaning of the name `plainFragment`?

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchClientTest.java
```diff
@@ -347,7 +556,38 @@ public void deliverResult_whenActivityIsNotABrowserSwitchListener_throwsIllegalA
...
+    @Test
+    public void convenience_deliverResultWithFragmentListener_forwardsInvocationToPrimaryDeliverResultMethod() {
+        when(fragmentAndListener.getActivity()).thenReturn(plainActivity);
+
+        sut = spy(BrowserSwitchClient.newInstance(browserSwitchConfig, activityFinder, persistentStore, returnUrlScheme));
+        doNothing().when(sut).deliverResult(any(FragmentActivity.class), any(BrowserSwitchListener.class));
+
+        sut.deliverResult(fragmentAndListener);
+        verify(sut).deliverResult(plainActivity, fragmentAndListener);
     }
```
scannillo, line 591: There's a lot in this test file. Can we organize it into sections by the methods being tested?

It looks like Android Studio supports labeled `region` comments that will code fold.

Ex:

```
//region deliverResult tests

    @Test 
    public void test_Method() { ... }

//endregion
```

scannillo:
I know this PR is triggered by a need in PPCP. Can you remind me of that use-case?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchClient.java
sshropshire: Good catch I think this was a formatting error.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchResult.java
sshropshire: Yeah `deliverResult()` currently won't produce an error. An error message can be produced when a `start()` method is called on `BrowserSwitchClient` and, for example, a merchant app is not properly configured for BrowserSwitch because `BrowserSwitchActivity` is missing in the AndroidManifest.xml.

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchClientTest.java
sshropshire: Yeah it's an androidx `Fragment` that doesn't implement the `BrowserSwitchListener` interface.

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchClientTest.java
sarahkoop: Added regions 👍

sshropshire:
For PPCP, when performing 3DS Validation via BrowserSwitch, we need access to the original `orderId` that initiated the 3DS authentication challenge. When we return to the app, there's no guarantee that this data will be available in the extreme case that the app is killed by the Android OS.

Since the `BrowserSwitchRequest` is persisted in SharedPrefs as a JSON string, we can add additional metadata to the request, and on `deliverResult()` we can retrieve the original `orderId`.

We'll probably end up using the new overload [here](https://github.com/paypal/android-sdk/blob/e70acf97097f0b135a773e3de94061679e7eb375/PayPal/src/main/java/com/paypal/androidsdk/CheckoutClient.java#L109).

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchClientTest.java
```diff
@@ -121,10 +171,16 @@ public void start_whenRequestCodeIsIntegerMinValue_notifiesListenerOfError() {
...
+        JSONObject metadata = new JSONObject();
+        BrowserSwitchOptions options = new BrowserSwitchOptions()
+                .requestCode(Integer.MIN_VALUE)
+                .url(uri)
+                .metadata(metadata);
+        sut.start(options, plainActivity, browserSwitchListener);
 
         ArgumentCaptor<BrowserSwitchResult> captor =
-            ArgumentCaptor.forClass(BrowserSwitchResult.class);
+                ArgumentCaptor.forClass(BrowserSwitchResult.class);
```
scannillo, line 183:
```suggestion
            ArgumentCaptor.forClass(BrowserSwitchResult.class);
```

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchClientTest.java
```diff
@@ -347,9 +566,48 @@ public void deliverResult_whenActivityIsNotABrowserSwitchListener_throwsIllegalA
...
-        sut.deliverResult(activity);
+        sut.deliverResult(plainActivity);
+    }
+
+    //endregion
+
+    //region test convenience deliverResult methods
+
+    @Test
+    public void convenience_deliverResultWithActivityListener_forwardsInvocationToPrimaryDeliverResultMethod() {
```
scannillo, line 577: I like this naming. It gets confusing with multiple overrides, so calling out the `forwardsInvocationToPrimaryMethod` is super clear 👍

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchOptions.java
scannillo: Woops - I meant the entire file needs docstrings, as well!

browser-switch/src/test/java/com/braintreepayments/browserswitch/BrowserSwitchRequestTest.java
```diff
@@ -38,5 +42,45 @@ public void toJson() throws JSONException {
         assertEquals(restored.getRequestCode(), original.getRequestCode());
         assertEquals(restored.getUri(), original.getUri());
         assertEquals(restored.getState(), original.getState());
+        assertNull(restored.getMetadata());
+    }
+
+    @Test
+    public void fromJson_withMetadata() throws JSONException {
+
```
scannillo, line 50:
```suggestion
```

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchPersistentStore.java
```diff
@@ -26,15 +31,21 @@ BrowserSwitchRequest getActiveRequest(Context context) {
         if (activeRequestJson != null) {
             try {
                 request = BrowserSwitchRequest.fromJson(activeRequestJson);
-            } catch (JSONException ignored) { /* do nothing */ }
+            } catch (JSONException e) {
+                Log.d(TAG, e.getMessage());
+                Log.d(TAG, Arrays.toString(e.getStackTrace()));
```
scannillo, line 36: Take it or leave it question: why logging here instead of having `getActiveRequest` throw an exception?

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchOptions.java
sshropshire: Done!

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchPersistentStore.java
sshropshire: Yeah this was mostly for internal purposes so we can catch errors locally during development. If we throw an exception we would have to bubble up to the merchant at some point via the API.

Since this class is non-public it's easier to just log the message for now and maybe we can re-consider in the future to implement more robust error handling for BrowserSwitch as a whole.

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchOptions.java
```diff
@@ -0,0 +1,89 @@
...
+     * @return {@link BrowserSwitchOptions} returns reference to instance to allow setter invocations to be chained
+     */
+    public BrowserSwitchOptions intent(Intent intent) {
+        this.intent = intent;
+        return this;
+    }
+
+    /**
+     * Set browser switch metadata
+     * @param metadata JSONObject containing metadata to associate with the browser switch request
```
sestevens, line 34: I wonder if we could expand this doc comment a bit to explain that this metadata property is intended for an information that you want to have available when returning from the browser. We could also state that it will be persisted even if the app is terminated during the browser switch.

```suggestion
     * @param metadata JSONObject containing metadata necessary for handling the return from browser to app. This data will be persisted and returned in {@link BrowserSwitchResult} even if the app is terminated during the browser switch.
```

sshropshire: Yeah I like that. It's good to explain the feature more.

PR #28, opened 2020-08-03
Gradle standardization

sarahkoop:
### Summary of changes

 - Reformat Gradle files to be more consistent and readable

 ### Checklist

 - [ ] ~Added a changelog entry~

### Authors

- @sshropshire @sarahkoop

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,19 @@ Add the library to your dependencies in your `build.gradle`:
 
 ```groovy
 dependencies {
-  compile 'com.braintreepayments:browser-switch:1.0.0'
+  implementation 'com.braintreepayments:browser-switch:1.0.0'
 }
 ```
 
-Declare `BrowserSwitchActivity` in your `AndroidManifest.xml`:
+To use the latest build from the `master` branch use:
+
+ ```groovy
+dependencies {
+  implementation 'com.braintreepayments:browser-switch:1.0.1-SNAPSHOT'
+}
+```
+
+Then, declare `BrowserSwitchActivity` in your `AndroidManifest.xml`:
 
 ```xml
 <activity android:name="com.braintreepayments.browserswitch.BrowserSwitchActivity"
--- a/deploy_snapshot.sh
+++ b/deploy_snapshot.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -e
+
+SLUG="braintree/android-card-form"
+JDK="oraclejdk8"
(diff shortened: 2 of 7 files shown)


deploy_snapshot.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -e
+
+SLUG="braintree/android-card-form"
```
sestevens, line 5: Is this supposed to be `browser-switch-android`?

README.md
```diff
@@ -12,11 +12,19 @@ Add the library to your dependencies in your `build.gradle`:
...
+  implementation 'com.braintreepayments:browser-switch:1.0.0'
 }
 ```
 
-Declare `BrowserSwitchActivity` in your `AndroidManifest.xml`:
+To use the latest build from the `master` branch use:
+
+ ```groovy
+dependencies {
+  implementation 'com.braintreepayments:browser-switch:1.0.1-SNAPSHOT'
```
sestevens, line 23: This will become out-of-date as we release new versions, since we're specifying `1.0.1-SNAPSHOT`. Is there a way to tell gradle to always pull in the latest snapshot?

sestevens: Or does the Rakefile increment this somehow?

deploy_snapshot.sh
sshropshire: 😅    yup haha. Missed this one in the copy paste. There was no `deploy_snapshot.sh` script originally.

README.md
sshropshire: Yeah this happens [here](https://github.com/braintree/browser-switch-android/blob/14b0bd23da08cd7636c9012eb6377f2736d9c0e7/Rakefile#L61). It would be nice to see if there are some official docs on naming SNAPSHOT artifacts. We advertise them in a lot of our repo's trying to make it more of a thing.

sestevens: Cool 👍

demerino: My only concern here would be offering multiple options to integrate. Is it a common practice to use SNAPSHOTs? In the past I think we've only used them as one-offs for merchants to test betas or potential fixes.

sshropshire: Yeah I understand that particular concern. Braintree core has advertised a snapshot for quite some time. We could eliminate snapshots across all projects that would make the README easier to read. I doubt a lot of merchants are using the snapshot builds since we never really get feedback from snapshot versions.

PR #29, opened 2020-08-03
Support custom tabs even when Chrome isn't installed

calvarez-ov:
### Summary of changes

Follow the example code in the [Chrome Custom Tabs documentation](https://developer.chrome.com/multidevice/android/customtabs) to see if any browser on the device supports custom tabs.

The previous behavior had a hardcoded check to only assume custom tabs support is available if the Chrome app is installed. Some users may only have the default browser, such as the Samsung browser, which supports custom tabs.

Note that the existing behavior, before this commit, actually already has partial support for custom tabs in other browsers: if Chrome is installed AND another compatible browser is installed, the other compatible browser is actually used: the user can choose it there's no default browser, or it's chosen automatically if it's the default browser.

This commit introduces consistent behavior for the checking for custom tabs support, and the actual opening of the url: alternative browsers are now considered for both, instead of just the latter.

Example scenarios:
==================

New behavior:
-------------
Only alternative browser:
* Have only Samsung internet installed
* Open the demo app
* Click on the "start browser switch" button
* Expected (new) behavior: browser switch url opened in custom tabs provided by Samsung Internet
* Actual (old) behavior: browser switch url opened in Samsung Internet browser without custom tabs

Unchanged behavior:
-------------------
Alternate browser is default:
* Have Chrome installed
* Have Samsung internet installed
* Have Samsung internet as the default browser
* Open the demo app
* Click on the "start browser switch" button
* Unchanged behavior: browser switch url opened in custom tabs provided by Samsung internet

Chrome is default:
* Have Chrome installed
* Have Samsung internet installed
* Have Chrome as the default browser
* Open the demo app
* Click on the "start browser switch" button
* Unchanged behavior: browser switch url opened in custom tabs provided by Chrome

Only Chrome:
* Have only Chrome installed
* Have Chrome as the default browser
* Open the demo app
* Click on the "start browser switch" button
* Unchanged behavior: browser switch url opened in custom tabs provided by Chrome

Chrome and alternate browser, no default:
* Have Chrome installed
* Have Samsung internet installed
* Have no default browser
* Open the demo app
* Click on the "start browser switch" button
* Unchanged behavior: User is prompted to choose a browser, and the url is opened in custom tabs provided by that browser

 ### Checklist

 - [x ] Added a changelog entry

- calvarez-ov

PR #30, opened 2020-08-04
Use the browser switch with custom tabs in the same activity task as `BraintreeFragment`'s activity

calvarez-ov:
### Summary of changes

Motivation:

When 3DS is done in an external browser inside another activity stack, sometimes the users try to navigate back to the app using back or recents:
* Either they navigate back to the app instead of browser after reading their 3DS code (in the sms app or bank app).
* Or, perhaps more rarely, the automatic deep link redirection after a successful code entry doesn't work, and users try to manually navigate back to the app using back or recents, instead of tapping the big blue "RETURN TO APP" button.

These scenarios result in the purchase being canceled.

This PR uses custom tabs in the same activity stack as the app. When the 3DS input succeeds, and the automatic redirection works, the custom tabs are automatically closed, and the calling activity (containing `BrowserSwitchFragment`) is brought to the top of the activity stack.

If the automatic redirection doesn't succeed, we still have the possible problem of a confused user who has to tap on the "RETURN TO APP" button. But since we're in the same activity stack as the app, the user can't navigate away, so the purchase can't be canceled.

Technical details:

* Don't add these flags when launching the browser intent: `FLAG_ACTIVITY_MULTIPLE_TASK`, `FLAG_ACTIVITY_CLEAR_TASK`, `FLAG_ACTIVITY_NEW_TASK`
* Launch the browser from the activity context, not from the application context (we can't use the application context anymore as that requires `FLAG_ACTIVITY_NEW_TASK`)
* Tell `BrowserSwitchActivity` which intent to launch when it finishes (`DemoActivity` in the demo example). This is done in two places:
  - The "normal" scenario is from `BrowserSwitchClient`, just before it launches custom tabs.
  - The other scenario is an edge case, in case the Android OS is aggressively killing our app process (reproducible by toggling "Don't keep activities" and "No background processes" in system developer settings). In this case, the application class tells `BrowserSwitchActivity` which intent to launch.

Note: I tried a few other approaches. For example, using an activity lifecycle callback listener to detect when `BrowserSwitchActivity` finishes, to relaunch `BrowserSwitchFragment`'s activity. These had various issues which can be found in the commit comments of the initial PR #18.

 ### Checklist

 - [x ] Added a changelog entry

- calvarez-ov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # browser-switch-android Release Notes
 
+## TBD
+
+* Use the browser switch with custom tabs in the same activity task as `BraintreeBrowserSwitch`s activity
+* Breaking changes:
+  * The activity which launches the browser switch must now have `android:launchMode="singleTask"`. See the `DemoActivity` entry in the [AndroidManifest.xml](demo/src/main/AndroidManifest.xml) file in the demo application
+  * To ensure that the calling activity will receive the the browser switch result even if the OS kills the app process when the browser is open, you should indicate the activity on app startup. See [DemoApplication](demo/src/main/java/com/braintreepayments/browserswitch/demo/DemoApplication.java) in the demo application:
+    ```java
+    BrowserSwitchActivity.setReturnIntent(new Intent(this, DemoActivity.class))
+    ```
+
 ## 1.1.3
 
 * Update androidx dependencies to latest versions
--- a/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
+++ b/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchActivity.java
@@ -1,8 +1,10 @@
 package com.braintreepayments.browserswitch;
 
 import android.app.Activity;
+import android.content.Intent;
 import android.os.Bundle;
 
+import androidx.annotation.NonNull;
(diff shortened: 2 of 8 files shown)


calvarez-ov:
This is the new version of #18, but based on version 1.0.0 of this library.

It's simpler than #18 as it doesn't include the following:
* persisting the return url: this was added in 1.0.0
* checking other compatible browsers: I move this to #29

sshropshire:
Hi @calvarez-ov thanks for using browser-switch. We see the concern here and we're planning on introducing something similar to what you're suggesting.

We have a new `BrowserSwitchOptions` object that may be able to help in this scenario. We'll post all updates here on GH.

calvarez-ov:
@sshropshire that sounds great! Looking forward to hearing more about it 😄

calvarez-ov:
Rebased onto master to have the latest changes from the 1.1.0 release.

calvarez-ov:
Rebased onto master to have the latest changes from the 1.1.3 release.

calvarez-ov:
Rebased onto master to have the latest changes from the 1.1.3 release.

sshropshire:
Hi @calvarez-ov we appreciate this PR. We took this into consideration and we've addressed the process kill issues in our [next major version](https://github.com/braintree/browser-switch-android/blob/master/CHANGELOG.md#200). Closing this for now.

PR #32, opened 2020-08-24
Fix BrowserSwitchFragment Null Pointer in `getReturnUrlScheme` method

sshropshire:
### Summary of changes

 - Fix BrowserSwitchFragment NPE when `getReturnUrlScheme` is called and an Activity is no longer attached to the fragment.

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop 
- @sshropshire

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
+++ b/browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
@@ -1,9 +1,11 @@
 package com.braintreepayments.browserswitch;
 
+import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


skoop:
Can I be removed as a mentioned user? I have nothing to do with this :)

sshropshire:
> Can I be removed as a mentioned user? I have nothing to do with this :)

Done! Also, shameless plug for our [PHP SDK](https://github.com/braintree/braintree_php) ✌️ 😄

browser-switch/src/main/java/com/braintreepayments/browserswitch/BrowserSwitchFragment.java
```diff
@@ -36,12 +48,7 @@ public void onResume() {
      */
     @SuppressWarnings("WeakerAccess")
```
hollabaq86, line 49: do we still need to suppress this warning?

sshropshire: Nope! Removing...

PR #36, opened 2021-02-01
Refactor Deliver Result 

sarahkoop:
### Summary of changes

 - Remove convenience `deliverResult` methods
 - Remove `BrowserSwitchListener`
 - Add activities for `launchMode:singleTop` and `launchMode:standard` to Demo Application

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop 
- @sshropshire


browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -91,57 +91,35 @@ private boolean isValidRequestCode(int requestCode) {
     }
 
     /**
-     * Deliver a pending browser switch result to an Android activity that is also a BrowserSwitchListener.
```
scannillo, line 94:
```
* Deliver a pending browser switch result to an Android activity
```

Don't we want to keep/modify this first sentence since it describes what the method actually does?

demo/src/main/java/com/braintreepayments/api/demo/DemoActivity.java
```diff
@@ -3,22 +3,71 @@
...
 
+        Bundle args = getIntent().getExtras();
+        if (args != null) {
+            returnUrlScheme = args.getString("returnUrlScheme");
+        }
+
+        browserSwitchClient = new BrowserSwitchClient();
+
         FragmentManager fm = getSupportFragmentManager();
         if (fm.findFragmentByTag(FRAGMENT_TAG) == null) {
```
sestevens, line 37: Would it make sense to use the `findDemoFragment` method here?
```suggestion
        if (findDemoFragment() == null) {
```

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
sshropshire: Yeah we can re-structure it a bit. The main reason for deletion here was because `BrowserSwitchListener` is no longer needed since the `deliverResult` method is now a synchronous call.

sestevens (approved):
Looks good to me!

demo/src/main/java/com/braintreepayments/api/demo/DemoActivity.java
sestevens: Also, this is a nit-pick, but it might make sense to call it `getDemoFragment` instead of `findDemoFragment`. You can take or leave that suggestion 🙂

scannillo (approved):
👍  🎊

PR #40, opened 2021-02-25
Check Return Url Scheme on Delivery

sarahkoop:
### Summary of changes

 - Fix issue of false successful result when `deepLinkUrl` did not match request `returnUrlScheme`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire 
- @sarahkoop

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/api/ChromeCustomTabsInternalClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/ChromeCustomTabsInternalClient.java
@@ -6,7 +6,6 @@ import android.net.Uri;
 import androidx.annotation.VisibleForTesting;
 import androidx.browser.customtabs.CustomTabsIntent;
 
-// TODO: unit test
 class ChromeCustomTabsInternalClient {
 
     private final CustomTabsIntent.Builder customTabsIntentBuilder;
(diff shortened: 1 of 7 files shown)


scannillo (approved):
This looks good to me - I am curious though, can you provide more details on how you surfaced the issue that led to this fix?

sshropshire:
> This looks good to me - I am curious though, can you provide more details on how you surfaced the issue that led to this fix?

Good question. We were integrating with popup bridge `beta1` and noticed an issue with `returnUrlScheme` and the `singleTop` Activity launch mode. Popup bridge also checks the return url scheme internally to make sure it matches a deep link to prevent a false positive "success" event. Overall, we felt like browser switch was a better location for the logic.

browser-switch/src/main/java/com/braintreepayments/api/ChromeCustomTabsInternalClient.java
```diff
@@ -0,0 +1,26 @@
+package com.braintreepayments.api;
+
+import android.content.Context;
+import android.net.Uri;
+
+import androidx.annotation.VisibleForTesting;
+import androidx.browser.customtabs.CustomTabsIntent;
+
+class ChromeCustomTabsInternalClient {
```
sshropshire, line 9: We added a "wrapper" around Chrome custom tabs to get Robolectric / Mockito to cooperate. We may consider this a best practice in the future for Android since it helps keep the tests clean by isolating our code from 3rd-party code.

sestevens (approved):

PR #46, opened 2021-08-09
Get Browser Switch Result

sshropshire:
### Summary of changes

 - Add `getResult()` method to `BrowserSwitchClient` to allow "previews" of a pending browser switch result before it is delivered.
 - Rename `BrowserSwitchClientTest` to `BrowserSwitchClientUnitTest`
 - Rename `ChromeCustomTabsInternalClientTest` to `ChromeCustomTabsInternalClientUnitTest`

### Notes
- `getResult()` method already has full test coverage through `deliverResult()`

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sshropshire

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -142,7 +142,7 @@ public class BrowserSwitchClient {
      *
      * @param activity the activity that received the deep link back into the app
      */
-    public BrowserSwitchResult getResult(@NonNull FragmentActivity activity) {
+    BrowserSwitchResult getResult(@NonNull FragmentActivity activity) {
         Intent intent = activity.getIntent();
         Context appContext = activity.getApplicationContext();
 
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -107,6 +107,42 @@ public class BrowserSwitchClient {
      * @param activity the activity that received the deep link back into the app
      */
     public BrowserSwitchResult deliverResult(@NonNull FragmentActivity activity) {
+        BrowserSwitchResult result = getResult(activity);
+        if (result != null) {
+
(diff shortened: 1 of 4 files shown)


browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -107,6 +107,42 @@ private boolean isValidRequestCode(int requestCode) {
...
+     * Peek at a pending browser switch result to an Android activity.
+     *
+     * We recommend you call this method in onResume to receive a browser switch result once your
+     * app has re-entered the foreground.
+     *
+     * This can be used in place of {@link BrowserSwitchClient#deliverResult(FragmentActivity)} when
+     * you want to know the contents of a pending browser switch result before it is delivered.
+     *
+     * @param activity the activity that received the deep link back into the app
+     */
```
scannillo, line 144: This is awesome - super clear

scannillo (approved):
LGTM

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -107,6 +107,42 @@ private boolean isValidRequestCode(int requestCode) {
...
+        return result;
+    }
+
+    /**
+     * Peek at a pending browser switch result to an Android activity.
+     *
+     * We recommend you call this method in onResume to receive a browser switch result once your
+     * app has re-entered the foreground.
+     *
+     * This can be used in place of {@link BrowserSwitchClient#deliverResult(FragmentActivity)} when
```
sarahkoop, line 140: I'm confused how this would work in a merchant app - they should call `getResult` and then also `deliverResult`? I'd opt to keep this method package-private unless we really think merchants need it because it introduces room for integration error in my opinion. If we need a CHANGELOG entry for the release we could just say `Added internal methods for Braintree Android Drop-in` or something.

sshropshire: Yeah I'm up for that. The CHANGELOG would've been my only reservation.

sarahkoop (approved):

PR #47, opened 2021-08-20
Android 12 Updates v1

sarahkoop:
### Summary of changes

 - Upgrade `compileSdkVersion` and `targetSdkVersion` to API 31 (Android 12)

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop


demo/src/main/AndroidManifest.xml
```diff
@@ -20,6 +21,7 @@
         <activity
             android:name="com.braintreepayments.browserswitch.BrowserSwitchActivity"
             android:launchMode="singleTask"
+            android:exported="true"
```
sshropshire, line 24: random q: does Android 12 require this?

sarahkoop: Yeah this is part of [Android 12 security updates](https://developer.android.com/about/versions/12/behavior-changes-12#exported)

sshropshire (approved):

demo/src/main/AndroidManifest.xml
sshropshire: Wow yeah that's legit. We may need to update the docs a bit.

demerino (approved):

PR #49, opened 2021-10-18
Fallback to Browser When Chrome Custom Tabs is Unavailable

sshropshire:
### Summary of changes

 - This PR checks that Chrome Custom Tabs is available before launching a Chrome Custom Tab.

 ### Checklist

 - [x] Added a changelog entry

### DIFF
--- a/browser-switch/src/main/AndroidManifest.xml
+++ b/browser-switch/src/main/AndroidManifest.xml
@@ -1,10 +1,14 @@
-<manifest package="com.braintreepayments.api.browserswitch"
-    xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.braintreepayments.api.browserswitch">
+
     <queries>
         <intent>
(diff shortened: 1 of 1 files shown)


scannillo:
Did we test that if this fallback occurs that the flow still switches back properly into the merchant app after it completes?

sshropshire:
> Did we test that if this fallback occurs that the flow still switches back properly into the merchant app after it completes?

Honestly, I haven't found an emulator yet that fully re-creates this scenario. I found one without chrome custom tabs that crashed because it had no WebView, but I think the emulator needs to have Google Play Services enabled. I'll have to keep trying all the way back to API 21.

sarahkoop:
Do we need to add this query intent to the manifest? [Source](https://developer.chrome.com/docs/android/custom-tabs/integration-guide/#how-can-i-check-whether-the-android-device-has-a-browser-that-supports-custom-tab)
```
<queries>
    <intent>
        <action android:name=
            "android.support.customtabs.action.CustomTabsService" />
    </intent>
</queries>
```

PR #50, opened 2021-12-20
Browser Switch Launcher and Observer

sshropshire:
This PR separates browser switch launch and browser switch result capture logic to allow for a cleaner separation of concerns. It also aims to make it clear that the browser switch initiating activity can be different from the browser switch deep link destination activity.

### Summary of changes

 - Extract `BrowserSwitchLauncher` from `BroswerSwitchClient`
 - Extract `BrowserSwitchObserver` from `BrowserSwitchClient`
 - Deprecate `BrowserSwitchClient`

 ### Checklist

 - [x] Added a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Create `BrowserSwitchLauncher`
 * Create `BrowserSwitchObserver`
-* Deprecate `BrowserSwitchClient`
+* Deprecate `BrowserSwitchClient` in favor of `BrowserSwitchLauncher`
 
 ## 2.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Create `BrowserSwitchLauncher`
 * Create `BrowserSwitchObserver`
-* Deprecate `BrowserSwitchClient` in favor of `BrowserSwitchLauncher`
+* Deprecate `BrowserSwitchClient` in favor of `BrowserSwitchLauncher` and `BrowserSwitchObserver`
 
 ## 2.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # browser-switch-android Release Notes
 
+## unreleased
+
+* Create `BrowserSwitchLauncher`
+* Create `BrowserSwitchObserver`
+* Deprecate `BrowserSwitchClient`
+
 ## 2.1.1
 
 * Fallback to browser when Chrome Custom Tabs is unavailable (thanks! @calvarez-ov)
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchLauncher.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchLauncher.java
@@ -31,7 +31,7 @@ public class BrowserSwitchLauncher {
         this.customTabsInternalClient = customTabsInternalClient;
     }
 
-    void launch(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException  {
+    public void launch(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException  {
         assertCanPerformBrowserSwitch(activity, browserSwitchOptions);
 
         Context appContext = activity.getApplicationContext();
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchLauncher.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchLauncher.java
@@ -1,0 +1,92 @@
+package com.braintreepayments.api;
+
+import android.content.Context;
+import android.content.Intent;
+import android.net.Uri;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
+import androidx.fragment.app.FragmentActivity;
+
+import org.json.JSONObject;
+
+public class BrowserSwitchLauncher {
+
+    private final BrowserSwitchInspector browserSwitchInspector;
+    private final BrowserSwitchPersistentStore persistentStore;
+
+    private final ChromeCustomTabsInternalClient customTabsInternalClient;
+
+    /**
+     * Construct a launcher to start a browser switch.
+     */
+    public BrowserSwitchLauncher() {
+        this(new BrowserSwitchInspector(), BrowserSwitchPersistentStore.getInstance(), new ChromeCustomTabsInternalClient());
+    }
+
+    @VisibleForTesting
+    BrowserSwitchLauncher(BrowserSwitchInspector browserSwitchInspector, BrowserSwitchPersistentStore persistentStore, ChromeCustomTabsInternalClient customTabsInternalClient) {
+        this.browserSwitchInspector = browserSwitchInspector;
+        this.persistentStore = persistentStore;
+        this.customTabsInternalClient = customTabsInternalClient;
+    }
+
+    void launch(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException  {
+        assertCanPerformBrowserSwitch(activity, browserSwitchOptions);
+
+        Context appContext = activity.getApplicationContext();
+
+        Uri browserSwitchUrl = browserSwitchOptions.getUrl();
+        int requestCode = browserSwitchOptions.getRequestCode();
+        String returnUrlScheme = browserSwitchOptions.getReturnUrlScheme();
+
+        JSONObject metadata = browserSwitchOptions.getMetadata();
+        BrowserSwitchRequest request =
+                new BrowserSwitchRequest(requestCode, browserSwitchUrl, metadata, returnUrlScheme, true);
+        persistentStore.putActiveRequest(request, appContext);
+
+        if (browserSwitchInspector.deviceHasChromeCustomTabs(appContext)) {
+            customTabsInternalClient.launchUrl(activity, browserSwitchUrl);
+        } else {
+            Intent launchUrlInBrowser = new Intent(Intent.ACTION_VIEW, browserSwitchUrl);
+            activity.startActivity(launchUrlInBrowser);
+        }
+    }
+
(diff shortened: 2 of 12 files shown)


demo/src/main/java/com/braintreepayments/api/demo/DemoFragment.java
```diff
@@ -14,14 +14,15 @@
...
+import com.braintreepayments.api.BrowserSwitchListener;
 import com.braintreepayments.api.BrowserSwitchOptions;
 import com.braintreepayments.api.BrowserSwitchResult;
 import com.braintreepayments.api.BrowserSwitchStatus;
 
 import org.json.JSONException;
 import org.json.JSONObject;
 
-public class DemoFragment extends Fragment implements View.OnClickListener {
+public class DemoFragment extends Fragment implements View.OnClickListener, BrowserSwitchListener {
```
sarahkoop, line 25: I think we should add `@Override` to the `onBrowserSwitchResult` method for this listener, but I can't comment on that line in this PR (line 104)

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # browser-switch-android Release Notes
 
+## unreleased
+
+* Create `BrowserSwitchLauncher`
+* Create `BrowserSwitchObserver`
+* Deprecate `BrowserSwitchClient`
```
sarahkoop, line 7:
```suggestion
* Deprecate `BrowserSwitchClient` in favor of `BrowserSwitchLauncher`
```

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchLauncher.java
```diff
@@ -0,0 +1,92 @@
...
+import android.content.Intent;
+import android.net.Uri;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
+import androidx.fragment.app.FragmentActivity;
+
+import org.json.JSONObject;
+
+public class BrowserSwitchLauncher {
```
sarahkoop, line 13:
```suggestion
/**
 * Class that manages the logic for browser switching.
 */
public class BrowserSwitchLauncher {
```

Not sure exactly what but should have a docstring

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchListener.java
```diff
@@ -0,0 +1,5 @@
+package com.braintreepayments.api;
+
+public interface BrowserSwitchListener {
```
sarahkoop, line 3: Docstring here too

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchLauncher.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @VisibleForTesting
+    BrowserSwitchLauncher(BrowserSwitchInspector browserSwitchInspector, BrowserSwitchPersistentStore persistentStore, ChromeCustomTabsInternalClient customTabsInternalClient) {
+        this.browserSwitchInspector = browserSwitchInspector;
+        this.persistentStore = persistentStore;
+        this.customTabsInternalClient = customTabsInternalClient;
+    }
+
+    public void launch(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException  {
```
sarahkoop, line 34:
```suggestion
   /**
     * Open a browser or <a href="https://developer.chrome.com/multidevice/android/customtabs">Chrome Custom Tab</a>
     * with a given set of {@link BrowserSwitchOptions} from an Android activity.
     *
     * @param activity the activity used to start browser switch
     * @param browserSwitchOptions {@link BrowserSwitchOptions} the options used to configure the browser switch
     */
    public void launch(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException  {
```

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchListenerFinder.java
```diff
@@ -0,0 +1,27 @@
...
+import androidx.fragment.app.FragmentActivity;
+import androidx.fragment.app.FragmentManager;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class BrowserSwitchListenerFinder {
+
+    List<BrowserSwitchListener> findActiveListeners(FragmentActivity activity) {
+        List<BrowserSwitchListener> result = new ArrayList<>();
```
sarahkoop, line 13: nitpick - can we call this list `listeners`?

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchObserver.java
```diff
@@ -0,0 +1,93 @@
...
+        BrowserSwitchResult result = getResult(activity);
+        if (result != null) {
+
+            Context appContext = activity.getApplicationContext();
+            BrowserSwitchRequest request = persistentStore.getActiveRequest(appContext);
+
+            List<BrowserSwitchListener> listeners = listenerFinder.findActiveListeners(activity);
+            for (BrowserSwitchListener listener : listeners) {
+                listener.onBrowserSwitchResult(result);
+            }
```
sarahkoop, line 40: Maybe this should all be consolidated - could we call `onBrowserSwitchResult` on the listeners as we are finding them? Might be unnecessary to loop through the list twice.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchObserver.java
```diff
@@ -0,0 +1,93 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
+import androidx.fragment.app.Fragment;
+import androidx.fragment.app.FragmentActivity;
+import androidx.fragment.app.FragmentManager;
+
+import java.util.List;
+
+public class BrowserSwitchObserver {
```
sarahkoop, line 15: Let's doc string this and the `onActivityResult` method too.

browser-switch/src/test/java/com/braintreepayments/api/BrowserSwitchListenerFinderUnitTest.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    private static class ListenerFragment extends Fragment implements BrowserSwitchListener {
+        @Override
+        public void onBrowserSwitchResult(BrowserSwitchResult result) {
+        }
+    }
+
+    private FragmentActivity plainActivity;
```
sarahkoop, line 33:
```suggestion
    private FragmentActivity activity;
```

browser-switch/src/test/java/com/braintreepayments/api/BrowserSwitchObserverUnitTest.java
```diff
@@ -0,0 +1,225 @@
...
+        ArgumentCaptor<BrowserSwitchRequest> requestCaptor = ArgumentCaptor.forClass(BrowserSwitchRequest.class);
+        verify(persistentStore).putActiveRequest(requestCaptor.capture(), same(activity));
+
+        BrowserSwitchRequest updatedRequest = requestCaptor.getValue();
+        assertSame(request, updatedRequest);
+        assertFalse(updatedRequest.getShouldNotifyCancellation());
+    }
+
+    @Test
+    public void onActivityResumed_whenDeepLinkUrlExistsAndReturnUrlSchemeDoesNotMatchAndShouldNotNotifyCancellation_returnsNull() {
```
sarahkoop, line 133:
```suggestion
    public void onActivityResumed_whenDeepLinkUrlExistsAndReturnUrlSchemeDoesNotMatchAndShouldNotNotifyCancellation_doesNothing() {
```

browser-switch/src/test/java/com/braintreepayments/api/BrowserSwitchObserverUnitTest.java
```diff
@@ -0,0 +1,225 @@
...
+        ArgumentCaptor<BrowserSwitchRequest> requestCaptor = ArgumentCaptor.forClass(BrowserSwitchRequest.class);
+        verify(persistentStore).putActiveRequest(requestCaptor.capture(), same(activity));
+
+        BrowserSwitchRequest updatedRequest = requestCaptor.getValue();
+        assertSame(request, updatedRequest);
+        assertFalse(updatedRequest.getShouldNotifyCancellation());
+    }
+
+    @Test
+    public void onActivityResumed_whenDeepLinkUrlDoesNotExistAndShouldNotNotifyCancellation_returnsNull() {
```
sarahkoop, line 192:
```suggestion
    public void onActivityResumed_whenDeepLinkUrlDoesNotExistAndShouldNotNotifyCancellation_doesNothing() {
```

sarahkoop:
Looks like we still have v1 integrations in the README - if we want to move forward with this we should probably update the README for it. If that's a separate PR that's ok with me too

sshropshire:
> Looks like we still have v1 integrations in the README - if we want to move forward with this we should probably update the README for it. If that's a separate PR that's ok with me too

Yeah that's a good call. We should update it can be this PR.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchObserver.java
sshropshire: It could be considered unnecessary but I think the tradeoff here makes `BrowserSwitchObserver` much easier to test since we don't have to think about `FragmentManager` or `instanceof`.

sshropshire: In the future, we could convert `BrowserSwitchListenerFinder` into an interface and let merchants define their own strategy for what's considered a listener, and have a `DefaultBrowserSwitchListenerFinder` implementation that will work for standard integrations.

browser-switch/src/test/java/com/braintreepayments/api/BrowserSwitchListenerFinderUnitTest.java
sshropshire: I agree the variable naming isn't clear here. I went with `nonListenerActivity` instead to show that it should skip notifying a `FragmentActivity` that doesn't implment `BrowserSwitchListener`.

sshropshire:
This design pattern is obsolete at the browser switch layer.

PR #53, opened 2022-09-02
Capture Browser Switch Result

sshropshire:
### Summary of changes

 - Add BrowserSwitchClient#captureResult() method to capture a browser switch result into persistent storage
 - Add BrowserSwitchClient#deliverResultFromCache() method to deliver a previously capture browser switch result 

 ### Checklist

 - [x] Added a changelog entry

- @jaxdesmarais 
- @sshropshire

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -169,4 +169,24 @@ public class BrowserSwitchClient {
 
         return result;
     }
+
+    BrowserSwitchResult getResultFromCache(@NonNull Context context) {
+        return persistentStore.getActiveResult(context.getApplicationContext());
+    }
+
+    public BrowserSwitchResult deliverResultFromCache(@NonNull Context context) {
+        Context applicationContext = context.getApplicationContext();
+        BrowserSwitchResult result = getResultFromCache(applicationContext);
+        if (result != null) {
+            persistentStore.clearActiveResult(applicationContext);
+        }
+        return result;
+    }
+
+    public void captureResult(@NonNull FragmentActivity activity) {
+        BrowserSwitchResult result = getResult(activity);
+        if (result != null) {
+            persistentStore.putActiveResult(result, activity.getApplicationContext());
+        }
+    }
 }
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -56,7 +56,8 @@ public class BrowserSwitchClient {
         persistentStore.putActiveRequest(request, appContext);
 
         if (browserSwitchInspector.deviceHasChromeCustomTabs(appContext)) {
-            customTabsInternalClient.launchUrl(activity, browserSwitchUrl);
+            boolean launchAsNewTask = browserSwitchOptions.isLaunchAsNewTask();
+            customTabsInternalClient.launchUrl(activity, browserSwitchUrl, launchAsNewTask);
         } else {
             Intent launchUrlInBrowser = new Intent(Intent.ACTION_VIEW, browserSwitchUrl);
             activity.startActivity(launchUrlInBrowser);
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -124,6 +124,7 @@ public class BrowserSwitchClient {
                 case BrowserSwitchStatus.SUCCESS:
                     // ensure that success result is delivered exactly once
                     persistentStore.clearActiveRequest(appContext);
+                    persistentStore.clearActiveResult(appContext);
                     break;
                 case BrowserSwitchStatus.CANCELED:
                     // ensure that cancellation result is delivered exactly once, but allow for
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # browser-switch-android Release Notes
 
+## unreleased
+
+* Add BrowserSwitchClient#captureResult() method to capture a browser switch result into persistent storage
+* Add BrowserSwitchClient#deliverResultFromCache() method to deliver a previously capture browser switch result
```
scannillo, line 6:
```suggestion
* Add BrowserSwitchClient#deliverResultFromCache() method to deliver a previously captured browser switch result 
```

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -168,4 +169,31 @@ BrowserSwitchResult getResult(@NonNull FragmentActivity activity) {
 
         return result;
     }
+
+    public BrowserSwitchResult deliverResultFromCache(@NonNull Context context) {
```
scannillo, line 173: Needs docs comment

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchPersistentStore.java
```diff
@@ -48,7 +51,35 @@ void putActiveRequest(BrowserSwitchRequest request, Context context) {
...
+    BrowserSwitchResult getActiveResult(Context context) {
+        BrowserSwitchResult request = null;
+
+        String activeResultJSON = PersistentStore.get(RESULT_KEY, context);
+        if (activeResultJSON != null) {
+            try {
+                request = BrowserSwitchResult.fromJson(activeResultJSON);
+            } catch (JSONException e) {
+                Log.d(TAG, e.getMessage());
+                Log.d(TAG, Arrays.toString(e.getStackTrace()));
```
scannillo, line 72: Would we rather return an error to the merchant instead of log?

sshropshire: Ideally yeah, the messed up part is it's been done this way historically for `putActiveRequest` too lol.

We do have an opportunity to make the code more clear in a next major version bump and reporting persistent storage errors may be one of the things we address down the line? There are also some alternatives to persistent storage that I've seen if we end up porting this library to Kotlin.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -168,4 +163,47 @@ BrowserSwitchResult getResult(@NonNull FragmentActivity activity) {
...
+    }
+
+    /**
+     * Capture a pending browser switch result for an Android activity.
+     * <p>
+     * To obtain the result in a separate activity, call {@link #deliverResultFromCache(Context)}.
+     *
+     * @param activity the activity that received the deep link back into the app
+     */
+    public void captureResult(@NonNull FragmentActivity activity) {
```
scannillo, line 192: Is `capture` a common term when using Android's `SharedPreferences`? If not, I think `persist` or `store` more clearly communicate the action here.

sshropshire: Yeah `put` would be the most direct alignment in terms of verbs.

Here's the invocation in the context of the [Core SDK PR](https://github.com/braintree/braintree_android/blob/269f207a4e603328a526ec75173715001bb59ec5/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeDeepLinkActivity.java#L15). I'd be down for `persist` or `store`. Capture comes from it being called in an invisible activity, but the naming doesn't have to align with Core.

scannillo: Gotcha. I don't feel too strongly, but wanted to ask. 

If we stick with `capture`, could we clarify in the docstring `* Capture a pending browser switch result for an Android activity into persistent storage.`?

sshropshire: Yeah definitely. That's a good call I'll update the docs string.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -8,6 +8,8 @@
 import androidx.annotation.VisibleForTesting;
 import androidx.fragment.app.FragmentActivity;
 
+import com.braintreepayments.api.browserswitch.R;
```
scannillo, line 11: What does this `.R` import do?

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchOptions.java
```diff
@@ -96,4 +98,13 @@ public Uri getUrl() {
...
     }
+
+    public boolean isLaunchAsNewTask() {
+        return launchAsNewTask;
+    }
+
+    public BrowserSwitchOptions launchAsNewTask(boolean launchAsNewTask) {
+        this.launchAsNewTask = launchAsNewTask;
+        return this;
+    }
```
scannillo, line 109: Should these just be a `get` and `set`?

scannillo (approved):
👏 Nice testing coverage for these changes

kgangineni (approved):
LGTM

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchOptions.java
sshropshire: We may keep this for now. We have a pseudo-builder pattern in place for this value object that removes the `set` prefix from all the mutating functions, and the `is` prefix is a common accessor naming convention when the underlying type is a boolean.

PR #56, opened 2022-12-13
Check if Activity is Finishing Before Browser Switch

sshropshire:
### Summary of changes

 - Fix issue that causes a browser switch to start while the host Activity is finishing.

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -57,7 +57,11 @@ public class BrowserSwitchClient {
                 new BrowserSwitchRequest(requestCode, browserSwitchUrl, metadata, returnUrlScheme, true);
         persistentStore.putActiveRequest(request, appContext);
 
-        if (browserSwitchInspector.deviceHasChromeCustomTabs(appContext)) {
+        if (activity.isFinishing()) {
+            String activityFinishingMessage =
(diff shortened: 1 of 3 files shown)


jaxdesmarais (approved):
👻

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -216,7 +220,7 @@ public void captureResult(@NonNull FragmentActivity activity) {
         Uri deepLinkUrl = intent.getData();
         if (deepLinkUrl != null) {
             BrowserSwitchResult result =
-                new BrowserSwitchResult(BrowserSwitchStatus.SUCCESS, request, deepLinkUrl);
+                    new BrowserSwitchResult(BrowserSwitchStatus.SUCCESS, request, deepLinkUrl);
```
mattwylder, line 223: Is this the indentation level we want?

mattwylder (approved):

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
sshropshire: Yeah it's the one enforced by the linter. It defaults to a double indent for continuation lines. We don't have a hard convention in place yet though might be worth changing.

PR #64, opened 2023-05-02
Parse Browser Switch Result

sshropshire:
### Summary of changes

* This PR takes an alternative approach to browser switching through the following methods
  * Add `BrowserSwitchClient#parseResult()` method
  * Add `BrowserSwitchClient#clearActiveRequests()` method
* Unlike `deliverResult()`, `parseResult()` does not mutate persistent storage
* `clearActiveRequests()` puts the responsibility on the caller to guard against delivery of a browser switch result more than one time 

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -176,7 +176,7 @@ public class BrowserSwitchClient {
      * @return
      */
     @Nullable
-    BrowserSwitchResult parseResult(@NonNull Context context, int requestCode, @Nullable Intent intent) {
+    public BrowserSwitchResult parseResult(@NonNull Context context, int requestCode, @Nullable Intent intent) {
         BrowserSwitchResult result = null;
         if (intent != null) {
             BrowserSwitchRequest request =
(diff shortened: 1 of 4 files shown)


browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -167,6 +167,40 @@ public BrowserSwitchResult getResult(@NonNull FragmentActivity activity) {
...
+    /**
+     * Parses and returns a browser switch result if a match is found.
+     *
+     * @param context     The context used to check for pending browser switch requests
+     * @param requestCode The request code for the matching pending request
+     * @param intent      Intent to evaluate for deep link result
+     * @return
+     */
+    @Nullable
+    public BrowserSwitchResult parseResult(@NonNull Context context, int requestCode, @Nullable Intent intent) {
```
scannillo, line 179: When should a merchant use `parseResult` over `deliverResult`?

scannillo: Gotcha, that would be helpful 

Maybe we can clarify in the `parseResult` docstring that it doesn't modify persistent storage in the same way as `deliver`

scannillo (approved):

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
sshropshire: Yeah I like that idea. I can update the docstring to call it out.

build.gradle
```diff
@@ -1,7 +1,7 @@
 buildscript {
     repositories {
+        mavenCentral()
         google()
-        jcenter()
```
sshropshire, line 4: JCenter repository has been EOL'd.

PR #69, opened 2023-08-29
Target Android 14

sarahkoop:
### Summary of changes

 - Upgrade compileSdkVersion and targetSdkVersion to API 34 (Android 14)
 - Update tests and fix lint errors to allow running test suite on later Java versions

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop


browser-switch/build.gradle
```diff
@@ -36,12 +36,6 @@ dependencies {
     testImplementation deps.mockitoCore
     testImplementation deps.jsonassert
     testImplementation deps.robolectric
-
-    // TODO: remove powermock
-    testImplementation 'org.powermock:powermock-module-junit4:2.0.9'
```
sarahkoop, line 41: This isn't directly related to Android 14 - but powermock was causing errors preventing me from running the test suite and we intended to remove it anyways

sshropshire: 🔥 Yeah this is awesome. This was also additional motivation for moving to Kotlin. Mockk is a lot more dev friendly.

sshropshire (approved):

browser-switch/src/test/java/com/braintreepayments/api/ChromeCustomTabsInternalClientUnitTest.java
```diff
@@ -1,38 +0,0 @@
...
-import androidx.browser.customtabs.CustomTabsIntent;
-
-import org.junit.Before;
-import org.junit.Test;
-
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
-
-public class ChromeCustomTabsInternalClientUnitTest {
```
sarahkoop, line 15: The mockito library upgrade allows us to mock static classes without powermock 🎉  but the started throwing errors mocking the final classes for CustomTabsIntent. I feel like we have sufficient test coverage of this flow (just the launching of the intent) through the integration tests, and we were previously mocking out the entirety of the functionality in the test, so I opted to just remove this test class in favor of integration tests. Thoughts?

scannillo (approved):

PR #76, opened 2023-12-05
Migrate FragmentActivity Parameters to ComponentActivity

sarahkoop:
### Summary of changes

 - Change all `FragmentActivity` parameters to `ComponentActivity`. This is more compatible with Jetpack Compose integrations, and we don't use any specific `FragmentActivity` features within the SDK, so we don't need to require it

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -4,10 +4,10 @@ import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
 
+import androidx.activity.ComponentActivity;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
-import androidx.fragment.app.FragmentActivity;
 
 import com.braintreepayments.api.browserswitch.R;
 
@@ -44,7 +44,7 @@ public class BrowserSwitchClient {
      * @param activity             the activity used to start browser switch
      * @param browserSwitchOptions {@link BrowserSwitchOptions} the options used to configure the browser switch
      */
-    public void start(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException {
+    public void start(@NonNull ComponentActivity activity, @NonNull BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException {
         assertCanPerformBrowserSwitch(activity, browserSwitchOptions);
 
         Context appContext = activity.getApplicationContext();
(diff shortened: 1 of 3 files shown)


browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -107,7 +107,7 @@ private boolean isValidRequestCode(int requestCode) {
      *
      * @param activity the activity that received the deep link back into the app
      */
-    public BrowserSwitchResult deliverResult(@NonNull FragmentActivity activity) {
+    public BrowserSwitchResult deliverResult(@NonNull ComponentActivity activity) {
```
sshropshire, line 110: Hey for BrowserSwitch `v3` we may may be able to get rid of this method in favor of `parseBrowserSwitchResult()`. The hope is to make the manual browser switching flow [mentioned in the migration guide](https://github.com/braintree/braintree_android/blob/main/v4.9.0%2B_MIGRATION_GUIDE.md#manual-browser-switching-for-browser-based-flows) the primary method of browser switching. It only needs a `Context` and an `Intent` instead of a particular `Activity` class.

sarahkoop: Ah ok yeah before `v3` I'll do some more cleanup to see if we even want to support this pattern (and we are already using `parseResult` in core now). But it's the `start` method that is still requiring activity. The only place the method actually needs an activity currently is to check if the activity is finishing and throw an error. This check may not be as necessary now that we are giving more control to merchants over invoking the method - do you think we can remove it and change the `start` parameter to context as well?

sshropshire: Oh yeah for `start()` I would definitely keep `ComponentActivity`. That's a really good point to make it Compose friendly.

It also makes sense to investigate turning Browser Switch into a launcher. That may help here, and apparently Compose has a [launcher API](https://developer.android.com/reference/kotlin/androidx/activity/compose/package-summary#rememberLauncherForActivityResult(androidx.activity.result.contract.ActivityResultContract,kotlin.Function1)) that should simplify `onCreate()` instantiation.

sshropshire: Only downside is we don't get a result back from Chrome Custom Tabs, but it does allow us to move away from `Activity.startActivity()`. It may be worth taking a look to see if it somehow makes sense for browser switching.

sshropshire: ^ there's a lot to discuss here, we have a ton of options. For the most part if we go with `parseResult()` and `start()` for now we should be good.

sshropshire (approved):

sarahkoop:
Replaced by #81

PR #77, opened 2023-12-12
Throw BrowserSwitchException When Unable to Start Activity

sarahkoop:
### Summary of changes

 - Catch the `ActivityNotFoundException` that is thrown in `BrowserSwitchClient#start` when a browser app is not available, and throw a `BrowserSwitchException` as defined by the method contract
 - Will be used to resolve [issue in core SDK](https://github.com/braintree/braintree_android/issues/799)

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -1,5 +1,6 @@
 package com.braintreepayments.api;
 
+import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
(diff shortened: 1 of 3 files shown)


browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -67,7 +68,11 @@ public void start(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOpti
             customTabsInternalClient.launchUrl(activity, browserSwitchUrl, launchAsNewTask);
         } else {
             Intent launchUrlInBrowser = new Intent(Intent.ACTION_VIEW, browserSwitchUrl);
-            activity.startActivity(launchUrlInBrowser);
+            try {
+                activity.startActivity(launchUrlInBrowser);
+            } catch (ActivityNotFoundException e) {
```
sarahkoop, line 73: Opted for this approach rather than a check if able to perform intent to avoid any false negatives on devices having browsers.

sshropshire (approved):

jaxdesmarais (approved):

PR #80, opened 2023-12-19
Publish Snapshots for v3 Branch

sarahkoop:
### Summary of changes

 - Publish snapshots for `v3` (next major version) branch to be used in core SDK major version development

 ### Checklist

 - ~[] Added a changelog entry~

### Authors

- @sarahkoop

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,7 +32,7 @@ plugins {
     id 'org.jetbrains.dokka' version '1.7.10'
 }
 
-version = '2.6.2-SNAPSHOT'
+version = '3.0.0-beta1-SNAPSHOT'
 group = "com.braintreepayments"
 ext {
     compileSdkVersion = 34
--- a/.github/workflows/release_snapshot.yml
+++ b/.github/workflows/release_snapshot.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - v3
 env:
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -32,7 +32,7 @@ plugins {
     id 'org.jetbrains.dokka' version '1.7.10'
 }
 
-version = '2.6.2-SNAPSHOT'
+version = '3.0.0-beta1-SNAPSHOT'
```
jaxdesmarais, line 35: Basic question that I am _sure_ it lives somewhere just not sure where, but I assume each release will increment this version?

jaxdesmarais (approved):

build.gradle
sarahkoop: This gets updated on actual releases [here](https://github.com/braintree/browser-switch-android/blob/67c237df2a488d10b1c6c035285a6cbf21501f3e/.github/workflows/release.yml#L86)and [here](https://github.com/braintree/browser-switch-android/blob/main/ci#L40), but not on snapshot releases - so this will remain `3.0.0-beta1-SNAPSHOT` until an actual release on this branch happens (which it won't - we will merge this into main before release)

saperi22 (approved):

scannillo (approved):

tdchow (approved):

PR #81, opened 2024-01-10
Add Integration Pattern without Local Storage

sarahkoop:
### Summary of changes

 - Add `BrowserSwitchClient.start` method that returns a `BrowserSwitchRequest` to allow integrators to store the pending request locally and pass it to new `BrowserSwithClient.parseResult` method to deliver browser switch results.
 - Remove storing pending requests in shared preferences in favor of allowing integrators to store the pending requests within their own app in whatever way works best for their architecture - this allows integrators to check if they have an existing pending request (i.e. they've launched a browser switch), before trying to fetch a result. The effect is that if they have a pending request, but don't receive a result, they can infer that the browser switch was cancelled or not completed and can handle accordingly. 
 - See this [draft PR](https://github.com/braintree/braintree_android/pull/883) for how this integration will look in the PayPal flow of the Braintree SDK
 - Future PRs will remove all other now unnecessary methods of BraintreeClient, and update the demo app

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -18,6 +18,7 @@ import org.json.JSONObject;
 /**
  * Client that manages the logic for browser switching.
  */
+// NEXT_MAJOR_VERSION remove all methods except start with ComponentActivity and parseResult with BrowserSwitchRequest
 public class BrowserSwitchClient {
 
     private final BrowserSwitchInspector browserSwitchInspector;
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -84,10 +84,9 @@ public class BrowserSwitchClient {
      *
      * @param activity the activity used to start browser switch
      * @param browserSwitchOptions {@link BrowserSwitchOptions} the options used to configure the browser switch
-     * @return a pending {@link BrowserSwitchRequest} that should be stored and passed to
-     * {@link BrowserSwitchClient#parseResult(BrowserSwitchRequest, Intent)} upon return to the app.
-     * Returns null if an exception is thrown.
-     * @throws BrowserSwitchException when a browser is unable to be launched for browser switching
+     * @return a {@link BrowserSwitchPendingRequest.Started} that should be stored and passed to
+     * {@link BrowserSwitchClient#parseResult(BrowserSwitchRequest, Intent)} upon return to the app,
+     * or {@link BrowserSwitchPendingRequest.Failure} if browser could not be launched.
      */
     @Nullable
     public BrowserSwitchPendingRequest start(@NonNull ComponentActivity activity, @NonNull BrowserSwitchOptions browserSwitchOptions) {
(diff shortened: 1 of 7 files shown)


browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
```diff
@@ -7,7 +7,7 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 
-class BrowserSwitchRequest {
+public class BrowserSwitchRequest {
```
sarahkoop, line 10: This needs to be public because it's now the return value of the `start` method - do we want a wrapper class for this? Like an object with Success/Failure values? Or is this ok?

browser-switch/src/main/java/com/braintreepayments/api/ChromeCustomTabsInternalClient.java
```diff
@@ -20,7 +21,7 @@ class ChromeCustomTabsInternalClient {
         this.customTabsIntentBuilder = builder;
     }
 
-    void launchUrl(Context context, Uri url, boolean launchAsNewTask) {
+    void launchUrl(Context context, Uri url, boolean launchAsNewTask) throws ActivityNotFoundException{
```
sarahkoop, line 24: This throws an exception if no browser is installed - ChromeCustomTabs intent just wraps plain browser intent so there's actually no need for a separate check for CCT capability - it will just fall back to non-CCT browser, or throw this exception if there's no browser at all.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
sshropshire: It's probably fine. From the SDK perspective, as on long as we mark browser-switch as `implementation` it won't be visible to the merchant unless they pull in browser switch on their own.

sshropshire: I would be up for renaming this to something like `BrowserSwitchPendingRequest` or `PendingBrowserSwitchRequest` though. Does that makes sense or is it too much?

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -76,7 +78,46 @@ public void start(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOpti
...
+                        new BrowserSwitchRequest(requestCode, browserSwitchUrl, metadata, returnUrlScheme, true);
+                customTabsInternalClient.launchUrl(activity, browserSwitchUrl, launchAsNewTask);
+            } catch (ActivityNotFoundException e) {
+                throw new BrowserSwitchException("Unable to start browser switch without a web browser.");
+            }
+            return request;
+        }
+    }
+
+    void assertCanPerformBrowserSwitch(ComponentActivity activity, BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException {
```
sshropshire, line 120: random: I wonder if we should make `assertCanPerformBrowserSwitch()` public as an explicit callout to Kotlin merchants since the `start()` method without a try/catch will still compile in Kotlin.

sshropshire (approved):

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
sarahkoop: Hmmm... it feels like a bit of unnecessary lift for the merchant - should we instead of throwing an error in the `start` method return a `BrowserSwitchPendingRequest.Success` or `BrowserSwitchPendingRequest.Failure` object? Similar to how we will wrap it in the BT SDK

sshropshire: Yeah that's a good call. Less API surface area too, so it's clear that it's the primary method for browser switching.

sarahkoop: Ok updated with the pending request type and removed throwing errors from this method all together. The next question will be how to parcelize/stringify the `BrowserSwitchPendingRequest` so that integrators can store it in whatever way they want. For our BT integration it doesn't matter but I've created a ticket to figure that out

PR #82, opened 2024-01-16
Update Demo App for v3 Integration

sarahkoop:
### Summary of changes

 - Update demo app to use new v3 integration pattern. Previous demo app had two activities (for standard or `singleTop` launch modes) both using a single fragment. To show different integration patterns, updated one of the activities to jetpack compose, and the other remains activity/fragment pattern
 - Convert `DemoActivity` to `ComposeActivity` and use jetpack compose. Also removed the start with metadata button option - this is just a param on browser switch options, so it doesn't really need it's entire own flow and tests. The default is just always to pass metadata now and have one single flow. 
 - Update `DemoActivitySingleTop` with shared code from old `DemoActivity` - This activity could definitely use more cleanup and modernization, but can be tackled in a future PR.
 - The UI doesn't entirely match up because I used the default compose UI for simplicity in that activity - I think it's ok that it doesn't match, but can update if we want it all to match. 

https://github.com/braintree/browser-switch-android/assets/67924275/3b991ccb-9dfa-468c-a3c9-410399380441

 ### Checklist

 - ~[ ] Added a changelog entry~

### Authors

- @sarahkoop


browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -47,36 +47,36 @@ public BrowserSwitchClient() {
...
-        } else {
-            Intent launchUrlInBrowser = new Intent(Intent.ACTION_VIEW, browserSwitchUrl);
-            try {
-                activity.startActivity(launchUrlInBrowser);
-            } catch (ActivityNotFoundException e) {
-                throw new BrowserSwitchException("Unable to start browser switch without a web browser.");
-            }
-        }
-    }
+//    public void start(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException {
```
sarahkoop, line 50: This will be removed in a future PR, just commented it out here because it was causing conflicts with the other `start` method in the demo app

build.gradle
```diff
@@ -30,6 +30,7 @@ buildscript {
 plugins {
     id 'io.github.gradle-nexus.publish-plugin' version '1.1.0'
     id 'org.jetbrains.dokka' version '1.7.10'
+    id 'org.jetbrains.kotlin.android' version '1.8.10' apply false
```
sarahkoop, line 33: The updates to gradle files were all auto-generated by the IDE when adding a compose activity. There is a separate ticket to update all gradle dependencies for browser switch v3 where this will be cleaned up and aligned with core dependencies

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
sshropshire: I'd vote for going ahead and removing this since we aren't gonna need it going forward. Is there a strong reason to keep?

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -88,7 +88,6 @@ public void start(@NonNull FragmentActivity activity, @NonNull BrowserSwitchOpti
      * {@link BrowserSwitchClient#parseResult(BrowserSwitchRequest, Intent)} upon return to the app,
      * or {@link BrowserSwitchPendingRequest.Failure} if browser could not be launched.
      */
-    @Nullable
```
sshropshire, line 91: Should this be `@NonNull` now?

demo/src/main/java/com/braintreepayments/api/demo/ComposeActivity.kt
```diff
@@ -0,0 +1,138 @@
...
+        Text(text = "Start Browser Switch")
+    }
+}
+
+@Composable
+fun BrowserSwitchSuccess(result: BrowserSwitchResult) {
+    val returnUrl = result.deepLinkUrl
+    returnUrl?.let {
+        val color = returnUrl.getQueryParameter("color")
+        val selectedColorString = String.format("Selected color: %s", color)
```
sshropshire, line 115:
```suggestion
        val selectedColorString = "Selected color: $color"
```

demo/src/main/java/com/braintreepayments/api/demo/ComposeActivity.kt
```diff
@@ -0,0 +1,138 @@
...
+        var metadataOutput: String? = null
+        result.requestMetadata?.let {
+            val metadataValue = result.requestMetadata!!.getString("test_key")
+            metadataOutput = "test_key=$metadataValue"
+        }
```
sshropshire, line 120:
```suggestion
        val metadataOutput = result.requestMetadata?.getString("test_key")?.let { "test_key=$it" }
```

demo/src/main/java/com/braintreepayments/api/demo/utils/PendingRequestUtil.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.braintreepayments.api.BrowserSwitchPendingRequest
+
+class PendingRequestUtil {
+
+    companion object {
+
+        private const val SHARED_PREFS_KEY = "PENDING_REQUESTS"
+        private const val PENDING_REQUEST_KEY = "BROWSER_SWITCH_REQUEST"
+
+        fun putPendingRequest(context: Context, pendingRequest: BrowserSwitchPendingRequest.Started) {
```
sshropshire, line 14:
```suggestion
        fun put(context: Context, pendingRequest: BrowserSwitchPendingRequest.Started) {
```
nit: `pendingRequestStore.put()` would read nicely in the optional unwrap chains.

demo/src/main/java/com/braintreepayments/api/demo/ComposeActivity.kt
```diff
@@ -0,0 +1,138 @@
...
+
+    override fun onResume() {
+        super.onResume()
+        PendingRequestUtil.getPendingRequest(this)?.let { startedRequest ->
+            val browserSwitchResult = browserSwitchClient.parseResult(startedRequest, intent)
+            browserSwitchResult?.let { result ->
+                viewModel.browserSwitchResult = result
+            } ?: run {
+                viewModel.browserSwitchError = Exception("User did not complete browser switch")
+            }
```
sshropshire, line 58: random: I wonder if we could be more explicit in the SDK response here. Returning `null` from `parseResult()` feels like the SDK is implicitly communicating to the merchant that the state of browser switching is unknown.

sarahkoop: Yeah @scannillo brought that up in the [PayPal PR too](https://github.com/braintree/braintree_android/pull/883)... I'm just not sure we want to be opinionated about the response here.

Really all the method is doing is checking if the pending request passed in has a matching result - since we are now leaving the storage of requests up to merchants, we can't really be certain if whatever request they pass in is in fact still pending. We could return an empty `BrowserSwitchResult` with status `UNKNOWN`, but that seems like more work for the merchant to unwrap than just returning null, since we can't really assert still pending when we don't know if the request that was passed in is in fact still pending. What do you think?

demo/src/main/java/com/braintreepayments/api/demo/utils/PendingRequestUtil.kt
sarahkoop: Done 👍

demo/src/main/java/com/braintreepayments/api/demo/utils/PendingRequestUtil.kt
```diff
@@ -0,0 +1,54 @@
+package com.braintreepayments.api.demo.utils
+
+import android.content.Context
+import android.content.SharedPreferences
+import com.braintreepayments.api.BrowserSwitchPendingRequest
+
+class PendingRequestUtil {
```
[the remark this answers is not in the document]
sarahkoop: Upated

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
sarahkoop: No there's just a separate ticket to remove all the unused methods - I can remove this one here just to keep it clean

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
sarahkoop: Yep 👍

demo/src/main/java/com/braintreepayments/api/demo/ComposeActivity.kt
sshropshire: My gut feeling is by making it non-null, we eliminate some uncertainty for merchants by giving them a documented enum in the case of failure.

Both solutions (`null` and `enum` results) will require merchants to include branching logic (i.e. if/else, when, let/run) in their apps to handle success and failure scenarios separately.

Enums fit well with `when(result) {}` and `if/else` in Kotlin. I do wish Kotlin had an equivalent to `if/let` in Swift, I feel like that would make `null` feel better. But still, my concern is that `null` doesn't fail loudly enough.

sarahkoop: Agreed. I think it's outside the scope of this demo app PR, but created a ticket to update from `null` to a result with `UNKNOWN` status

sshropshire (approved):

PR #85, opened 2024-01-24
Upgrade Gradle Dependencies

sarahkoop:
### Summary of changes

 - Upgrade gradle dependencies to latest (and align with BT Core SDK versions)
 - Bump Java version to 11 and Kotlin to 1.9.10
 - Upgrade to Android Gradle Plugin 8 (aligns with core)
 - Bump min SDK version to 23 - there isn't anything specific in this SDK that requires the min version bump, but this gives us more flexibility and aligns with the Core v5 min SDK version

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/browser-switch/build.gradle
+++ b/browser-switch/build.gradle
@@ -5,7 +5,8 @@ plugins {
 }
 
 android {
-    compileSdkVersion rootProject.compileSdkVersion
+    namespace "com.braintreepayments.api.browserswitch"
+    compileSdk rootProject.compileSdkVersion
 
     defaultConfig {
         minSdkVersion rootProject.minSdkVersion
@@ -15,8 +16,11 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility = 1.8
-        targetCompatibility = 1.8
+        sourceCompatibility versions.javaSourceCompatibility
+        targetCompatibility versions.javaTargetCompatibility
+    }
+    kotlinOptions {
+        jvmTarget = "11"
     }
 
     // robolectric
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -7,37 +7,44 @@ buildscript {
         }
     }
 
+    def sdkTargetJavaVersion = JavaVersion.VERSION_11
+    ext.versions = [
+            "javaSourceCompatibility": sdkTargetJavaVersion,
+            "javaTargetCompatibility": sdkTargetJavaVersion,
+    ]
+
```
jaxdesmarais, line 15: nit: looks like there may be an extra space here

jaxdesmarais (approved):

browser-switch/build.gradle
```diff
@@ -5,7 +5,8 @@ plugins {
 }
 
 android {
-    compileSdkVersion rootProject.compileSdkVersion
+    namespace "com.braintreepayments.api.browserswitch"
```
sshropshire, line 8: random: This might be a good time to migrate the actual code packages to `com.braintreepayments.api.browserswitch`. It'll lessen the probability of collision with other libraries. Definitely work for a separate PR.

Package private has worked well for us, but it may make the Core SDK more robust if we migrate to `@RestrictTo` annotations one day. Could also be `v6` work. Bringing it up now so we don't forget.

sshropshire (approved):

browser-switch/build.gradle
sarahkoop: Discussed and have a ticket to investigate this further 👍

build.gradle
sarahkoop: Fixed

PR #90, opened 2024-04-02
[V3] Encode Pending Browser Switch Request as Base64 String

sshropshire:
### Summary of changes

 - Rename `BrowserSwitchPendingRequest` to `BrowserSwitchStartResult`
 - Support `ComponentActivity` for `BrowserSwitchClient#start()` method
 - Rename `BrowserSwitchResult` to `BrowserSwitchParseResult`
 - Remove `BrowserSwitchInfo` helper class

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
```diff
@@ -48,24 +66,20 @@ JSONObject getMetadata() {
...
 
-    String toJson() throws JSONException {
-        JSONObject result = new JSONObject();
-        result.put("requestCode", requestCode);
-        result.put("url", url.toString());
-        result.put("returnUrlScheme", returnUrlScheme);
-        result.put("shouldNotify", shouldNotifyCancellation);
-        if (metadata != null) {
-            result.put("metadata", metadata);
+            byte[] requestJSONBytes = requestJSON.toString().getBytes(StandardCharsets.UTF_8);
```
sshropshire, line 78: Base 64 encoding is technically optional, though it does make the state string appear to the merchant as an [opaque data type](https://en.wikipedia.org/wiki/Opaque_data_type) to discourage tampering. Base64 encoding also has an added benefit of [masking the data](https://security.stackexchange.com/questions/203664/whats-the-purpose-of-base64-if-knowing-its-base64-makes-it-easy-to-decode) to prevent casual observation of browser switch metadata e.g. intercepting an ec-token over a screen share.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -106,23 +105,30 @@ private boolean isValidRequestCode(int requestCode) {
...
-     * {@link BrowserSwitchClient#start(ComponentActivity, BrowserSwitchOptions)}
-     * @param intent the intent to return to your application containing a deep link result from the
-     *               browser flow
-     * @return a {@link BrowserSwitchResult.Success} if the browser switch was successfully
-     * completed, or {@link BrowserSwitchResult.NoResult} if no result can be found for the given
-     * {@link BrowserSwitchPendingRequest.Started}. A {@link BrowserSwitchResult.NoResult} will be
+     *
+     * @param intent              the intent to return to your application containing a deep link result from the
+     *                            browser flow
+     * @param pendingRequestState the {@link BrowserSwitchStartResult.Success} token returned from
```
sarahkoop, line 111: I'm not sure about the state/token naming here, those don't align in my mind. Maybe `startResultDetails` or `startResultInfo` or `storedStartResult` and remove `token` from the description?

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
```diff
@@ -1,39 +1,58 @@
...
-        JSONObject jsonObject = new JSONObject(json);
-        int requestCode = jsonObject.getInt("requestCode");
-        String url = jsonObject.getString("url");
-        String returnUrlScheme = jsonObject.getString("returnUrlScheme");
-        JSONObject metadata = jsonObject.optJSONObject("metadata");
-        boolean shouldNotify = jsonObject.optBoolean("shouldNotify", true);
-        return new BrowserSwitchRequest(requestCode, Uri.parse(url), metadata, returnUrlScheme, shouldNotify);
+
+    @NonNull
+    static BrowserSwitchRequest fromBase64EncodedJSON(@NonNull String base64EncodedRequest) throws BrowserSwitchException {
```
sarahkoop, line 35: I guess we should weigh the benefits of base64encoding compared to the extra merchant integration lift of having to handle the `Failure` type for all browser based flows. I could go either way. What do others think?

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
sshropshire: I agree the naming can be improved. Choosing the right names for Browser Switch will maximize the Developer Experience and influence the DX in core.

For me the only thing with "details" and "info" that's conflicting is an implication that the string we return can provide information at runtime. Technically it's an [opaque string](https://stackoverflow.com/a/5838979) that we don't want merchants to inspect.

We should definitely remove the word token, that's a holdover from what I called it originally. But in a way this "state" is similar to a BT nonce–we don't expect merchants to inspect the value for details because the structure of the data is undefined i.e. doesn't follow semantic versioning.

Could we maybe shorten it to `requestState`? Or continue to ideate as a group this might be something to get some 👀's on. I'd be interested in seeing how this bubbles up to the features in the Core SDK too that could help us pick the right names.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
sshropshire: Oh yeah def. We can't really enforce JSON tampering though. We may still need the failure scenario for the case where we're unable to reconstruct the original `BrowserSwitchRequest`.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
tdchow: I think `storedStartResult` makes sense. "result" seems generic enough for a merchant to not want to inspect its value.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
```diff
@@ -1,39 +1,58 @@
...
 
+import java.nio.charset.StandardCharsets;
+
+// Links
+// Base64 Encode a String in Android: https://stackoverflow.com/a/7360440
+
+// TODO: consider encryption
+// Ref: https://medium.com/fw-engineering/sharedpreferences-and-android-keystore-c4eac3373ac7
+
+// TODO: Rename to `BrowserSwitchStartRequest` and remove `BrowserSwitchOptions` in favor this class
```
tdchow, line 20: Should we remove these TODOs before merging the PR in?

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
sshropshire: I like `storedStartResult` for this method. The only thing where the naming doesn't scale as nicely is when the result is [unpacked](https://github.com/braintree/browser-switch-android/blob/3b8bf6256dde8580966094763d1ac8503714ee71/demo/src/main/java/com/braintreepayments/api/demo/ComposeActivity.kt#L74), we'd end up with `result.storedStartResult`.

I'm wondering now if we should just go with something like `originalRequest` to keep it simple? It'll still be opaque because the type is `String`, and it would work nicely in both contexts (`start()` and `parse()`).

Also reiterating the main reason we make this value a `String` is because strings work nicely with all Android persistence mechanisms e.g. SharedPrefs, DataStore, etc. We only need the merchant to hold on to this value for us to remove the responsibility of state restoration out of the SDK. We've gotten a good number of inbounds on that in the past, which makes us think that state restoration logic may be unique to each individual merchant app.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
sshropshire: I agree, we may be able to convert these to JIRA tickets.

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
tdchow: The first `result` would be the What if we had separate naming for the `start()` and `parse()` steps? It might give a better description for each for each step. 

Start - `startResultToStore`
``` kotlin
        when (val result = browserSwitchClient.start(this, browserSwitchOptions)) {
            is BrowserSwitchStartResult.Success ->
                PendingRequestStore.put(this, result.startResultToStore)
            ...
        }
```

Parse - `storedStartResult`
```kotlin
       PendingRequestStore.get(this)?.let { storedStartRequest ->
            when (val result = browserSwitchClient.parseResult(intent, storedStartResult)) {
                ...
            }
        }
```

sarahkoop: Honestly I still kind of think the `pendingRequest` naming is a little more clear for what it is - a request that hasn't been completed yet. Maybe we could change the types to `Started` and `Failure` to better indicate that the request hasn't completed successfully yet? 

```
   when (val pendingRequest = browserSwitchClient.start(this, browserSwitchOptions)) {
            is BrowserSwitchPendingRequest.Started ->
                PendingRequestStore.put(this, pendingRequest.requestToStore)
            ...
        }
```

Also maybe something like `completeRequest` would make more sense than `parse` to indicate what is required by the merchant?

```
       PendingRequestStore.get(this)?.let { storedPendingRequest ->
            when (val result = browserSwitchClient.completeRequest(intent, storedPendingRequest)) {
                ...
            }
        }
```

Can't decide what would be most clear for merchants. Tagging in @scannillo @jaxdesmarais @saperi22 for more opinions

tdchow: PR for renaming `parseResult()` to `completeRequest()`: https://github.com/braintree/browser-switch-android/pull/101

sarahkoop:
This PR might need to be closed and re-worked after all the changes in browser switch - @sshropshire can you revert the naming changes or open a new PR with just the base64 encoding work?

sshropshire:
Closing this after discussion with @sarahkoop. Will create a new PR with the Base64 encoding changes.

PR #96, opened 2024-05-08
Demo App - Add App Link toggle to DemoFragment

tdchow:
### Summary of changes

 - Add deep link / app link toggle to `DemoFragment`
 - Configure app link flow using `https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com` url

Screenshot_20240508_161515

Related popup bridge PR: https://github.com/braintree/popup-bridge-example/pull/10

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/demo/src/main/java/com/braintreepayments/api/demo/DemoFragment.java
+++ b/demo/src/main/java/com/braintreepayments/api/demo/DemoFragment.java
@@ -5,7 +5,10 @@ import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.AdapterView;
+import android.widget.ArrayAdapter;
 import android.widget.Button;
+import android.widget.Spinner;
 import android.widget.TextView;
 
 import androidx.annotation.IdRes;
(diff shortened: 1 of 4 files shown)


sarahkoop (approved):

demo/src/main/java/com/braintreepayments/api/demo/DemoFragment.java
```diff
@@ -79,10 +101,14 @@ private void startBrowserSwitch(@Nullable JSONObject metadata) {
         }
     }
 
-    private Uri buildBrowserSwitchUrl() {
+    private Uri buildBrowserSwitchUrl(boolean isAppLink) {
         String url = "https://braintree.github.io/popup-bridge-example/" +
-                "this_launches_in_popup.html?popupBridgeReturnUrlPrefix=" +
-                getReturnUrlScheme() + "://";
+            "this_launches_in_popup.html?";
```
richherrera, line 106: qq: why do we need to concatenate these two strings?

richherrera (approved):

demo/src/main/java/com/braintreepayments/api/demo/DemoFragment.java
tdchow: I think the single string might have reached the character line limit. I thought it'd be more readable to keep the `"this_launches_in_popup.html?"` portion together.

PR #98, opened 2024-05-10
Handle when BrowserSwitchResult has a null returnUrlScheme and update matchesAppLinkUri logic

tdchow:
### Summary of changes

 - Handle when `BrowserSwitchResult` has a null `returnUrlScheme` for the App Link flow
 - Update `matchesAppLinkUri` logic to only check for a matching host and scheme

https://github.com/braintree/browser-switch-android/assets/5005216/54645fba-acaf-4ed8-9305-e0b89e6733fd

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
@@ -22,11 +22,14 @@ class BrowserSwitchRequest {
         JSONObject jsonObject = new JSONObject(json);
         int requestCode = jsonObject.getInt("requestCode");
         String url = jsonObject.getString("url");
-        String returnUrlScheme = jsonObject.getString("returnUrlScheme");
         JSONObject metadata = jsonObject.optJSONObject("metadata");
         Uri appLinkUri = null;
         if (jsonObject.has("appLinkUri")) {
             appLinkUri = Uri.parse(jsonObject.getString("appLinkUri"));
+        }
+        String returnUrlScheme = null;
+        if (jsonObject.has("returnUrlScheme")) {
+            returnUrlScheme = jsonObject.getString("returnUrlScheme");
         }
         boolean shouldNotify = jsonObject.optBoolean("shouldNotify", true);
         return new BrowserSwitchRequest(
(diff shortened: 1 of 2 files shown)


saperi22 (approved):

richherrera (approved):

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
```diff
@@ -22,12 +22,15 @@ static BrowserSwitchRequest fromJson(String json) throws JSONException {
...
-        String returnUrlScheme = jsonObject.getString("returnUrlScheme");
         JSONObject metadata = jsonObject.optJSONObject("metadata");
         Uri appLinkUri = null;
         if (jsonObject.has("appLinkUri")) {
             appLinkUri = Uri.parse(jsonObject.getString("appLinkUri"));
         }
+        String returnUrlScheme = null;
+        if (jsonObject.has("returnUrlScheme")) {
+            returnUrlScheme = jsonObject.getString("returnUrlScheme");
+        }
```
sshropshire, line 33: nit: `optString` may also work here.
```suggestion
        String returnUrlScheme = jsonObject.optString("returnUrlScheme");
```

sshropshire (approved):

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchRequest.java
tdchow: I wasn't aware of `optString()`! I looked through the docs for the function. It looks like the default value when the `name` isn't found is an empty string, not null 😞 It might not affect anything, but I think setting null properties on the `BrowserSwitchResult` is clearer than empty strings. 

```
    /**
     * Returns the value mapped by {@code name} if it exists, coercing it if
     * necessary, or the empty string if no such mapping exists.
     */
    @NonNull public String optString(@Nullable String name) {
        return optString(name, "");
    }

    /**
     * Returns the value mapped by {@code name} if it exists, coercing it if
     * necessary, or {@code fallback} if no such mapping exists.
     */
    @NonNull public String optString(@Nullable String name, @NonNull String fallback) {
        Object object = opt(name);
        String result = JSON.toString(object);
        return result != null ? result : fallback;
    }
```

sshropshire: Yeah that's cool.

sshropshire: It's ones and zeroes at the end of the day. The way you had it is more expressive though. The biggest lift with `JSONObject` has always been avoiding exceptions.

PR #99, opened 2024-05-21
Visibility Changes for BT Venmo Module

tdchow:

### Summary of changes

 - Update assertCanPerformBrowserSwitch to a public method
 - Update gradle for enabling local maven publishing
 - Bump version to 3.0.0-beta2-SNAPSHOT

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+* Make `BrowserSwitchClient.assertCanPerformBrowserSwitch()` public
+
 * Breaking Changes
   * Bump `minSdkVersion` to API 23
   * Bump target Java version to Java 11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## unreleased
 
 * Make `BrowserSwitchClient.assertCanPerformBrowserSwitch()` public
-
 * Breaking Changes
   * Bump `minSdkVersion` to API 23
   * Bump target Java version to Java 11
--- a/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
+++ b/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
@@ -79,7 +79,17 @@ public class BrowserSwitchClient {
         }
     }
 
-    void assertCanPerformBrowserSwitch(ComponentActivity activity, BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException {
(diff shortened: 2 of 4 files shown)


saperi22 (approved):

browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java
```diff
@@ -79,7 +79,17 @@ public BrowserSwitchPendingRequest start(@NonNull ComponentActivity activity, @N
...
 
-    void assertCanPerformBrowserSwitch(ComponentActivity activity, BrowserSwitchOptions browserSwitchOptions) throws BrowserSwitchException {
+    /**
+     * Throws a {@link BrowserSwitchException} when a browser switch flow cannot be started.
+     *
+     * @param activity the activity used to start browser switch
+     * @param browserSwitchOptions {@link BrowserSwitchOptions} the options used to configure the browser switch
+     * @throws BrowserSwitchException exception containing the error message on why browser switch cannot be started
+     */
+    public void assertCanPerformBrowserSwitch(
```
sarahkoop, line 89: Is restricting to library scope similar to what we are doing in the core SDK in v5 not possible here?

tdchow: Unfortunately, the restrict to annotation wouldn't work here as it'd be tied to the Gradle group ID which is different between browser switch and Venmo. The PayPal module also uses this function.

It also might make sense to make public to other consumers of Browser Switch. Since we're using the function within the BT SDK, other consumers might want to access it as well.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+* Make `BrowserSwitchClient.assertCanPerformBrowserSwitch()` public
+
```
sarahkoop, line 6:
```suggestion
* Make `BrowserSwitchClient.assertCanPerformBrowserSwitch()` public
```

sarahkoop (approved):

CHANGELOG.md
tdchow: I'm not seeing any differences in the suggested change. Can you update the snippet?

sarahkoop: Oh I was trying to remove the newline after the entry - nitpick it's fine either way

tdchow: Ah I see. I just removed the newline 👍

PR #109, opened 2024-07-09
Add Temp Post Release workflow

tdchow:

### Summary of changes

 - Add a temporary workflow to run the `bump_version` and `create_github_release` jobs which were not run for the 3.0.0-beta1 release

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,6 +8,7 @@ on:
       job_to_run:
         description: 'Run individual job (DO NOT USE THIS FOR A NORMAL RELEASE)'
         required: false
+        type: choice
         options:
           - bump_version
           - create_github_release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,13 +5,6 @@ on:
       version:
         description: 'Version to release'
         required: true
-      job_to_run:
-        description: 'Run individual job (DO NOT USE THIS FOR A NORMAL RELEASE)'
-        required: false
-        type: choice
-        options:
-          - bump_version
-          - create_github_release
 env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -5,6 +5,13 @@ on:
       version:
         description: 'Version to release'
         required: true
+      job_to_run:
+        description: 'Run individual job (DO NOT USE THIS FOR A NORMAL RELEASE)'
+        required: false
+        type: choice
+        options:
+          - bump_version
+          - create_github_release
```
sarahkoop, line 14: Should we maybe put this in a separate work flow instead to avoid any release process confusion? Like release docs or something?

tdchow: I thought about that, but we'd need to either define the jobs twice (in both workflows) and keep them in sync, or we have 2 separate workflows that we'll need to run for a given release.

Not great in both cases 🙁 maybe we can remove this change once the docs and github release are completed.

sarahkoop: I'm also slightly confused how `job_to_run` maps to actually invoking the `options` - is there any documentation for that param?

tdchow: Shoot i'm missing the ifs inside each job to check for the field, but I realize that will end up breaking running the entire workflow.

Do you suggest creating a separate workflow for these 2 jobs, then removing it after they've completed?

sarahkoop: Hmm yeah that seems like probably the best options

sshropshire (approved):

jaxdesmarais (approved):

PR #121, opened 2025-10-07
Completed AuthTab Implementation

noguier:

### Summary of changes
- added AuthTabInternalClient, that support both AuthTab and ChromeCustomTab.
Note: ACTIVITY_NEW_TASK is not supported for AuthTab
- removed ChromeCustomTabsInternalClient,
- modified ComposeActivity & DemoActivitySingleTop to accept the above changes. Added checks to use Authtab or ChromeCustomTab depending on the device
- added toast visuals to see whether or not the AuthTab is supported
- cleared intent data after handling incomplete browser switch to prevent displaying stale success data from a previous request

 ### Checklist

 - [ ] Added a changelog entry

- @noguier


tdchow:
I was thinking of potential ways to streamline the integration steps for Browser Switch with Auth Tab. 

What if we removed the exposed `AuthTabCallback` used for the Auth Tab flow, but internally set the `AuthTabIntent.registerActivityResultLauncher`. When the launcher callback is invoked with a result, we could store it in the `BrowserSwitchClient` class and use that result when the caller of the SDK invokes `BrowserSwitchClient.completeRequest()`, instead of the passed in intent.

This way the caller of the Browser Switch SDK does not need to add additional logic for AuthTab and can use the existing API interface.

@noguier @saperi22 - I haven't completely vetted this approach but wanted to get your thoughts!

PR #135, opened 2026-04-27
Create instrumentation_tests.yml

buzzamus:

### Summary of changes
- Created GitHub Actions Instrumentation Tests workflow to run instrumentation tests in CI

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Had Claude use the instrumentation tests workflow yaml file from Braintree Android repo to create similar workflow for this repo's instrumentation tests

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/.github/workflows/instrumentation_tests.yml
+++ b/.github/workflows/instrumentation_tests.yml
@@ -7,7 +7,15 @@ jobs:
       fail-fast: false
       matrix:
         api-level: [ 31, 35 ]
-        target: [ default ]
+        target: [ default, google_apis ]
+        profile: [ pixel_7_pro ]
+        include:
+          - api-level: 23
+            target: default
+            profile: Nexus 6
+          - api-level: 23
+            target: google_apis
+            profile: Nexus 6
     steps:
       - name: checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


noguier:
I see you removed the some APIs, there were no way to make the 23 and google apis work? Just curious

buzzamus:
> I see you removed the some APIs, there were no way to make the 23 and google apis work? Just curious

I was able to get the other Google API version numbers to work, but API 23 I believe is having difficulty discovering UI elements due to differences in the older API version

noguier:
I think we are good to go without api 23 then, thank you!

.github/workflows/instrumentation_tests.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: retry tests
+        if: steps.run_tests.outcome == 'failure'
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: ${{ matrix.api-level }}
+          target: ${{ matrix.target }}
+          arch: x86_64
+          profile: pixel_7_pro
+          script: |
+            adb shell 'echo "_ --disable-fre --no-default-browser-check --no-first-run" > /data/local/tmp/chrome-command-line'
```
noguier, line 44: do you mind giving me some context why we needed to add shell commands to our running script, I am just curious.

noguier (approved):

.github/workflows/instrumentation_tests.yml
buzzamus: Yep! Of course: This is to get around the first time Chrome is launched bring up the "Terms of Service" popup that blocks the text the test is looking for, so it essentially launches Chrome as if it has already been launched before

noguier: ahh, smart!

saralvasquez (approved):
Looks good and also looks like the run and retry are working as expected. I think it's good to go in!

PR #136, opened 2026-04-28
Add CLAUDE.md with build commands and architecture overview

saperi22:
### Summary of changes
- Add CLAUDE.md

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

-

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -50,5 +50,5 @@ Published to Maven Central via Sonatype OSSRH. Requires env vars `SONATYPE_NEXUS
 
 ### Versions
 
+- Current major version: v3.x; v2.x deprecated March 2026
 - compileSdk 36, minSdk 23, Java 11, Kotlin 1.9.20
-- Current: `3.5.2-SNAPSHOT`; v2.x deprecated March 2026
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,54 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Commands
+
+```bash
+./gradlew build                        # Full build with tests
+./gradlew test                         # Unit tests (Robolectric, no device needed)
+./gradlew test --tests "*ClassName*"   # Run a single test class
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,54 @@
...
+Tests use **Robolectric** (no emulator required), JUnit 4, Mockito, and MockK. All tests live in `browser-switch/src/test/java/`. `BrowserSwitchClientUnitTest.java` is the primary test file covering the main client behavior.
+
+### Publishing
+
+Published to Maven Central via Sonatype OSSRH. Requires env vars `SONATYPE_NEXUS_USERNAME`, `SONATYPE_NEXUS_PASSWORD`, and signing keys (`SIGNING_KEY_ID`, `SIGNING_KEY_PASSWORD`, `SIGNING_KEY_FILE`). Release tasks: `changeGradleReleaseVersion`, `updateCHANGELOGVersion`, `incrementSNAPSHOTVersion`.
+
+### Versions
+
+- compileSdk 36, minSdk 23, Java 11, Kotlin 1.9.20
+- Current: `3.5.2-SNAPSHOT`; v2.x deprecated March 2026
```
saralvasquez: Do we want the current version to be an unreleased snapshot or the current released version according to the CHANGELOG. Also will this get updated on release?

saralvasquez (commented):
Small question but otherwise looks good

CLAUDE.md
saperi22: This doesn't get updated. I could just remove this part. I don't think this file needs to be the source on which version is the latest.
We could easily get it from the release tab/ tags.

saralvasquez (approved):
🎉

jaxdesmarais (approved):
Thanks for adding this! The key classes was really helpful for me as someone that hasn't worked on browser switch much to get an overview of what it does. 😄

PR #139, opened 2026-07-28
feat: add android 17 instrumentation test support

saralvasquez:
### Summary of changes
- adds android 17 to the instrumentation test matrix

Adding a CHANGELOG entry isn't necessary since this is only a CI change. It also doesn't change any functionality of the repo

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
it wasn't

- [ ] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @saralvasquez

### DIFF
--- a/.github/workflows/instrumentation_tests.yml
+++ b/.github/workflows/instrumentation_tests.yml
@@ -3,11 +3,22 @@ on: [pull_request, workflow_dispatch]
 jobs:
   test:
     runs-on: ubuntu-latest
+    env:
+      EMULATOR_OPTIONS: -no-window -gpu swiftshader_indirect -dns-server 8.8.8.8 -noaudio -no-boot-anim -camera-back none -memory 4096 -feature GLDirectMem,HasSharedSlotsHostMemoryAllocator
     strategy:
       fail-fast: false
       matrix:
-        api-level: [ 31, 35 ]
-        target: [ default, google_apis ]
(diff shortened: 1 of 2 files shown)


.github/workflows/instrumentation_tests.yml
```diff
@@ -3,11 +3,22 @@ on: [pull_request, workflow_dispatch]
...
+          - api-level: 31
+            target: default
+          - api-level: 31
+            target: google_apis
+          - api-level: 35
+            target: default
+          - api-level: 35
+            target: google_apis
+          - api-level: "37.0"
+            target: google_apis_ps16k
```
mzlangreder, line 21: Ah this is nice, do you think we can apply this same style to our other integration tests so we can deduplicate the sections we added for Android 17 in https://github.com/braintree/braintree_android/ ?

mzlangreder (approved):

agedd (approved):

.github/workflows/instrumentation_tests.yml
saralvasquez: Yeah I think we probably can. I'll make a ticket to investigate that

